PR #40, opened 2014-11-03
Align KATCPSensor listener callback API with katcp.Sensor

brickZA:
also:

Improve KATCPSensor listener callback mechanism to properly deal with methods

@ludwigschwardt Please comment also :)

### DIFF
--- a/katcp/resource.py
+++ b/katcp/resource.py
@@ -2,10 +2,12 @@
 
 import abc
 import collections
+import logging
 
 from katcp import Message
 from katcp.sampling import SampleStrategy
 
+logger = logging.getLogger(__name__)
 
 def normalize_strategy_parameters(params):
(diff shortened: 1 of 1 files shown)


brickZA:
@ludwigschwardt My philosophy here is to try and have things be the same between the basic katcp stuff and the KATCPResource stuff as far as it makes sense. What I like about the current version is that code written for a normal katcp.Sensor object would also work as a listner for a KATCPSensor object, so long as it is using the named attributes (the KATCPSensor ones are a strict superset). 

I do also understand where you are coming from. Apart from changing the KATCPSensor API as you suggest, would you be happy if we add a computed "status_str: attribute to the named tuple? Of course this is potentially incompatible with existing code that expects the string, but AFAIK the string version of the callback was a relatively new addition. Or perhaps we need to talk philosophy IRL :)

PR #57, opened 2015-01-23
Add resource groups

brickZA:


lvdheever:
Looks good to me. I'll merge unless someone else is reviewing it.
I like the use of "groups" instead of "arrays"

Is there is any merit in keeping this and "the old way" in sync for a while as we go along so that we can compare old and new implementations easily. E.g. change
    [katconn:arrays]
    ants = ${system.resource_names('res:receptor')}
to [groups]. 
Or are you close enough to completion to just activate it all in one? 
What do you think?

brickZA:
@lvdheever 

The implementation is fairly removed from config, so we can do the config change at any time.

For the API that is presented to observations the problem is that @ludwigschwardt is unavailable to work on it till after some conference, but that should be fairly easy to align with him when he's back.

PR #59, opened 2015-01-26
Added initial_state to AsyncState and explicitly set initial state of InspectingClient

theunsa:
@brickZA: to merge ... Added initial_state to AsyncState and explicitly set initial state of InspectingClient to InspectingClientStateType with all attributes False


brickZA:
I'm just adding some tests

PR #61, opened 2015-01-30
Fake resource support

brickZA:
Functionality for clients that fake server connections on the client side. Useful for testing, and hopefully useful for kattelmod's time-warping telescope simulation, since all clients share the same ioloop allowing ioloop time-warp tricks to be used. Allows clients at the level op InspectingClient, KATCPClientResource and KATCPClientResourceContainer to be faked, with the ability for request-handlers and sensors to be installed on the 'fake' server.

@ludwigschwardt  FYI

### DIFF
--- a/katcp/fake_clients.py
+++ b/katcp/fake_clients.py
@@ -1,5 +1,7 @@
 # Copyright 2015 SKA South Africa (http://ska.ac.za/)
 # BSD license - see COPYING for details
+
+
 import tornado.concurrent
 
 from thread import get_ident as get_thread_ident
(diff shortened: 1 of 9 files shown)


katcp/fake_clients.py
```diff
@@ -0,0 +1,408 @@
...
+        """
+        rh_dict = {}
+        for name in dir(rh_obj):
+            if not callable(getattr(rh_obj, name)):
+                continue
+            if name.startswith("request_"):
+                request_name = convert_method_name("request_", name)
+                req_meth = getattr(rh_obj, name)
+                # Check that all the handlers have docstrings as strings
+                assert req_meth.__doc__, "Even fake request handlers must have docstrings"
```
martinslabber: This will later also be assert-ed by add_request_handlers_dict

katcp/fake_clients.py
```diff
@@ -0,0 +1,408 @@
...
+        self.child_managers[child_name].add_sensors(sensor_infos)
+
+    def add_request_handlers_object(self, child_name, rh_obj):
+        """Add fake request handlers from an object with request_* method(s)
+
+        See :class:`FakeInspectingClientManager` for detail
+        """
+        return self.child_managers[child_name].add_request_handlers_object(rh_obj)
+
+    def add_request_handlers_dict(self, rh_dict):
```
martinslabber: child_name should be part of method args.

PR #71, opened 2015-02-24
Resolve broken decorator part of CAM-425

brickZA:


katcp/kattypes.py
```diff
@@ -693,15 +709,39 @@ def raw_handler(self, *args):
...
+    Major version of KATCP to use when interpreting types.
+    Defaults to latest implemented KATCP version.
+
+
+Examples
+--------
+>>> class MyDeviceClient(katcp.client.AsyncClient):
+...     @inform(Int(), Float())
+...     def inform_myinf(self, req, my_int, my_float):
+...         '''Handle #myinf <y_int> <my_float> inform received from server'''
```
martinslabber: I guess <y_int> should be <my_int>

brickZA: Yes, and also removed incorrect 'req' parameter

PR #72, opened 2015-02-25
Fix docstring example error, remove redundant comment

brickZA:


katcp/kattypes.py
```diff
@@ -707,8 +707,6 @@ def raw_handler(self, msg):
 
     return decorator
 
-# Using partial with no extra parameters lets us 'copy' the function so that we
-# can change the docstring without affecting the original function's docstring
```
theunsa: Why remove this useful comment?

brickZA: Because it is no longer true -- the partial call is also applying a
parameter, which makes it the normal use of partial, and therefore not
comment-worthy.

theunsa: Ok, but I guess modifying it to remove the "with no extra parameters" part would make it true. The comment-worthiness for me is the 'docstring override' part.

But anyway, no biggy.

brickZA: Actually, point taken, see #73 :)

PR #74, opened 2015-03-02
Add basic KATCPSensensor.wait()

brickZA:
Also implement KATCPResource.wait() in usingKATCPSensor.wait()

### DIFF
--- a/katcp/resource.py
+++ b/katcp/resource.py
@@ -710,6 +710,7 @@ class KATCPSensor(object):
                 'Cannot wait on a sensor that does not have a strategy set')
 
         def handle_update(sensor, reading):
+            # This handler is called whenever a sensor update is received
             try:
                 assert sensor is self
                 val_matched = reading.value == condition_or_value
@@ -727,6 +728,7 @@ class KATCPSensor(object):
                 self.unregister_listener(handle_update)
 
         self.register_listener(handle_update)
+        # Handle case where sensor is already at the desired value
         ioloop.add_callback(handle_update, self, self._reading)
 
         if timeout:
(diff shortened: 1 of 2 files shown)


katcp/resource.py
```diff
@@ -690,6 +669,72 @@ def get_status(self):
...
+                    ioloop.add_callback(f.set_result, True)
+            except Exception:
+                f.set_exc_info(sys.exc_info())
+                self.unregister_listener(handle_update)
+
+        self.register_listener(handle_update)
+        ioloop.add_callback(handle_update, self, self._reading)
+
+        if timeout:
+            to = ioloop.time()+timeout
```
theunsa: pep8 ... spaces around '+'

katcp/resource.py
```diff
@@ -690,6 +669,72 @@ def get_status(self):
...
+                status_matched = reading.status == status or status is None
+                if val_matched and status_matched:
+                    self.unregister_listener(handle_update)
+                    ioloop.add_callback(f.set_result, True)
+            except Exception:
+                f.set_exc_info(sys.exc_info())
+                self.unregister_listener(handle_update)
+
+        self.register_listener(handle_update)
+        ioloop.add_callback(handle_update, self, self._reading)
```
theunsa: Question? The ioloop.add_callback line will result in 'handle_update' being called on the next ioloop iteration ... so it is not to say that the sensor had an updated value yet. Where is the actual delay to allow the conditions to be met within 'timeout' time?

theunsa: Question2? What happens if the sensor gets an updated value before the next ioloop iteration gets to call 'handle_update'?

katcp/resource.py
brickZA: Done

PR #75, opened 2015-03-04
Format bool Message argument as '1'/'0' rather than 'True'/'False'

brickZA:


katcp/core.py
```diff
@@ -291,6 +290,14 @@ def __init__(self, mtype, name, arguments=None, mid=None):
                                    "alphabetic character (got %r)."
                                    % (name,))
 
+    def format_argument(self, arg):
+        if isinstance(arg, float):
```
martinslabber: Doc string please. Even just a short little one.

PR #78, opened 2015-03-10
Add support for KATCPResource request activation / deactivation

brickZA:


katcp/resource.py
```diff
@@ -350,8 +383,27 @@ def __call__(self, *args, **kwargs):
...
+    @abc.abstractmethod
+    def issue_request(self, *args, **kwargs):
+        """Signature as for __call__
+
+        Do the request immediately without checking active state.
         """
 
+    def is_active(self):
+        """True if resource for this request is active"""
+        return self._is_active()
```
theunsa: `self._is_active` is not a method

brickZA: Indeed not a method, but it is a callable member that is set by __init__() unless I'm missing something.

The design-idea is that KATCPRequest doesn't itself have any idea of how 'activity' is determined, but trusts the callable passed in at __init__ time.

PR #85, opened 2015-04-09
Add off level

bxaia:
@martinslabber please can you review and merge.

- OFF is the highest possible logging level and is intended to turn logging off.
  No information. Devices should never log messages directly to the OFF logging level.

### DIFF
--- a/katcp/server.py
+++ b/katcp/server.py
@@ -2264,7 +2264,7 @@ class DeviceLogger(object):
         WARN: logging.WARN,
         ERROR: logging.ERROR,
         FATAL: logging.FATAL,
-        OFF: logging.CRITICAL + 10 # OFF is the highest possible logging level
+        OFF: logging.FATAL + 10 # OFF is the highest possible logging level
     }
 
     def __init__(self, device_server, root_logger="root", python_logger=None):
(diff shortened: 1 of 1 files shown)


katcp/server.py
```diff
@@ -2325,7 +2326,7 @@ def set_log_level(self, level):
         """
         self._log_level = level
         if self._python_logger:
-            level = PYTHON_LEVEL.get(level)
+            level = self.PYTHON_LEVEL.get(level)
```
martinslabber: can we make this more safe:

Where you say: 
 level = self.PYTHON_LEVEL.get(level)
please change that to:
 level = self.PYTHON_LEVEL.get(level, logging.NOTSET)

So if an strange log level is requested KATCP will not blow up but will just log everything, since it has no filter. An example is the log level ALL as defined in the list but not implemented in the dict.

katcp/server.py
```diff
@@ -2264,6 +2264,7 @@ class DeviceLogger(object):
         WARN: logging.WARN,
         ERROR: logging.ERROR,
         FATAL: logging.FATAL,
+        OFF: logging.CRITICAL + 10 # OFF is the highest possible logging level
```
martinslabber: logging.CRITICAL and logging.FATAL are the same thing. But it would make more sense in this file if you stick to the one or the other. since FATAL is used every where rather reference that here.
OFF: logging.FATAL + 10

katcp/server.py
theunsa: Good point, but I'd beware of the setting it to NOTSET ... in such a case you will flood your logging system unintentionally and probably unknowingly. 

Rather just revert back to the default which is WARN.

theunsa: On second thought ... you could just index using [ ] and then catch the ValueError and report a more proper error message so that you know why KATCP blows up. That is how `def set_log_level_by_name(self, level_name)` does it.

PR #87, opened 2015-04-23
Add host and port

profcalculus:


katcp/resource_client.py
```diff
@@ -262,6 +270,12 @@ def until_state(self, state):
         """
         return self._state.until_state(state)
 
+    def wait_connected(self, timeout=None):
+        """Future that resolves when the state is not 'disconnected'
+        TODO: implement timeout
```
brickZA: I don't think timeout should be added here -- The blocking interface could perhaps add a timeout parameter that will only wait for the future to resolve until that timeout is reached, otherwise the normal tornado.gen.with_timeout or something similiar  can be used.

PR #88, opened 2015-04-30
Lize add is connected

lvdheever:
Added is_connected for backwards compatibility 
and changed "Failed to connect" log from warn to debug


katcp/resource_client.py
```diff
@@ -259,6 +259,13 @@ def __init__(self, resource_spec, parent=None, logger=log):
         self._sensors_synced = AsyncCallbackEvent(self._update_state)
         self._requests_synced = AsyncCallbackEvent(self._update_state)
 
+    def is_connected(self):
+        """Indication of the connection state
+
+            Desired state, one of ("disconnected", "syncing", "synced")
```
profcalculus: Slightly confusing docstring.
How about:
""" Connection state.
Returns True if the state is 'syncing' or 'synced'.
"""

PR #92, opened 2015-05-07
Allow for wrapping of attributes of type `resource.ClientGroup`

theunsa:
@lvdheever / @profcalculus 
Allow for wrapping of attributes of type `resource.ClientGroup` (i.e. to ensure the group requests are blocking when wrapped)


theunsa:
Note that this wraps the requests but nothing else, e.g:

kat.ants.req.sensor_list() will be wrapped to be blocking 
kat.ants.clients[0].req.sensor_list() will still return a future (same for kat.ants.clients[0].sensor.blah)

PR #108, opened 2015-05-15
Add resource client to resource container

fjoubert:
@theunsa or @lvdheever 

Please review and merge.


katcp/resource_client.py
```diff
@@ -909,6 +909,18 @@ def set_sensor_listener(self, sensor_name, listener):
                 child_sensor_name = sensor_name[len(prefix):]
                 child.set_sensor_listener(child_sensor_name, listener)
 
+    def add_child_resource_client(self, res_name, res_spec):
```
brickZA: Perhaps document that the child resource must not yet bet started, otherwise there might be ioloop mixing madness going on.

PR #112, opened 2015-06-18
Extend log message to include reading and type

ludwigschwardt:


ludwigschwardt:
This will help to debug those slews of sensor update errors in the data proxy logs.

PR #117, opened 2015-07-22
Adding phython identifier to KATCPSensor object

theunsa:


katcp/resource.py
```diff
@@ -595,10 +595,15 @@ def parent_name(self):
 
     @property
     def name(self):
-        """Name of this KATCP resource"""
+        """Name of this KATCPSensor"""
         return self._name
 
     @property
+    def python_id(self):
```
martinslabber: I am happy with the code.
Personally I do not like the method name python_id maybe normalised_name
@theunsa and I chatted we have a hand full of possible names.
 We think @lvdheever can decide on a nice name.

brickZA: A little verbose perhaps, but 'python_identifier_name' says exactly what it is. i.e. a name to use when a valid python-identifier is needed. Official python description of what is allowed:

https://docs.python.org/2/reference/lexical_analysis.html#grammar-token-identifier

theunsa: @brickZA: Yeah, I did actually consider that exact name but when talking to @martinslabber I'm realised that the sensor names are also used outside of the context of python (i.e. when a client subscribes to a sensor through katportal)

theunsa: @martinslabber also said that for katstore query they always normalise the name they get so it does not matter what the client provides ... we will probably do the same thing in katportal. I should create a doodle ... ha.

lvdheever: Hmm - no specific preference. But may be "normalised_name" to make it less python specific. Or what about something very different like "sensor_handle" or "sensor_unih" for "unique handle". Or is it just not good to introduce such new terminology now?

brickZA: I like 'normalised_name'. You could also call it 'slug', which is the term
used by many web content management systems  when turning a title into an
URL (and also used in the newspaper industry for a similar purpose, which
is where it came from)

Cheers
Neilen

martinslabber: Another reason to make it "normalise_name" would be that we need to address the sensor name with underscores outside the python code eg. in gui or in the hdf5 files. And I already call it normalise_name in katstore ;)

lvdheever: OK - so shall we go with 'normalised_name' then. Seems like there are 3 half-votes for that one.

PR #122, opened 2015-09-01
Lize fix tests

lvdheever:
@theunsa Please review carefully. I think there were some potential problems in the Container set_sensor_sampling and set_sensor_listener and I reworked both.

Also moved the two failing tests (test_set_sensor_sampling and test_set_sensor_listener) to the integrated container test class so that the test covers more and use less mocks.  Please also review the tests to see that it correctly test those two functions.

### DIFF
--- a/katcp/resource_client.py
+++ b/katcp/resource_client.py
@@ -881,39 +881,37 @@ class KATCPClientResourceContainer(resource.KATCPResource):
     def set_sensor_strategy(self, sensor_name, strategy_and_parms):
         sensor_name = resource.escape_name(sensor_name)
         sensor_obj = getattr(self.sensor, sensor_name, None)
-        print "====",sensor_name,strategy_and_parms, sensor_obj
-        if not sensor_obj:
-            print self.sensor,sensor_name
-        child_name = sensor_obj.parent_name
-        child = self.children[child_name]
-        print "++++",sensor_obj,child_name ###,sensor_obj.parent_name
         if sensor_obj:
-                # Handle aggregate sensors that are not prefixed with "parent_name_"
-                if sensor_name.startswith("agg_"):
-                    print "----A",sensor_name,strategy_and_parms
-                    yield child.set_sensor_strategy(sensor_name, strategy_and_parms)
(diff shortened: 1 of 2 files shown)


katcp/resource_client.py
```diff
@@ -881,33 +881,37 @@ def list_sensors(self, filter="", strategy=False, status="",
...
+        # Otherwise, depend on self._add_sensors() to handle it when the sensor appears
 
     @steal_docstring_from(resource.KATCPResource.set_sensor_listener)
     def set_sensor_listener(self, sensor_name, listener):
         sensor_name = resource.escape_name(sensor_name)
         sensor_obj = getattr(self.sensor, sensor_name, None)
-        for child_name in dict.keys(self.children):
+        if sensor_obj:
+            # The sensor exists, so register the listener.
+            child_name = sensor_obj.parent_name
```
ludwigschwardt: The parent is a child too? :-)

lvdheever: Yes - the parent of sensor_obj is a child of the container.
Will see if there is some rework that can be done on the naming.

katcp/resource_client.py
```diff
@@ -881,33 +881,37 @@ def list_sensors(self, filter="", strategy=False, status="",
     def set_sensor_strategy(self, sensor_name, strategy_and_parms):
         sensor_name = resource.escape_name(sensor_name)
         sensor_obj = getattr(self.sensor, sensor_name, None)
-        for child_name in dict.keys(self.children):
+        if sensor_obj:
+            # The sensor exists, so set the strategy.
+            child_name = sensor_obj.parent_name
```
theunsa: I agree with @ludwigschwardt that we should distinguish who's-child are we manipulating here.

theunsa:
Rest is fine.

theunsa:
@lvdheever: Just a ping ... if we call that variable something more accurate I'll merge ... does not necessarily have to happen everywhere ... at least for the new code.

lvdheever:
Renaming done.

PR #123, opened 2015-09-07
Lize fix tests

lvdheever:
The change is to specify the resource_name for set_sensor_listener and set_sampling_strategy on the Container (so that it knows on which child resource to cache/set the sensor listener/strategy)

### DIFF
--- a/katcp/resource_client.py
+++ b/katcp/resource_client.py
@@ -881,57 +881,65 @@ class KATCPClientResourceContainer(resource.KATCPResource):
     def set_sensor_strategy(self, sensor_name, strategy_and_parms):
         sensor_name = resource.escape_name(sensor_name)
         sensor_obj = getattr(self.sensor, sensor_name, None)
-        for child_name in dict.keys(self.children):
-            child = self.children[child_name]
-            if sensor_name.startswith("agg_"):
-                if sensor_obj and (child.name == sensor_obj.parent_name):
-                    # Handle aggregate sensors that are not prefixed with "parent_name_"
-                    yield child.set_sensor_strategy(sensor_name, strategy_and_parms)
-                    break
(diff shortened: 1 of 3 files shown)


theunsa:
Looks good ... I'll merge but it might have ripple effect where `set_sensor_strategy` is used.

katcp/resource_client.py
```diff
@@ -391,8 +391,22 @@ def list_sensors(self, filter="", strategy=False, status="",
             dict.items(self.sensor), filter, strategy, status, use_python_identifiers, tuple, refresh)
 
     @tornado.gen.coroutine
-    @steal_docstring_from(resource.KATCPResource.set_sensor_strategy)
```
brickZA: I would prefer it if you kept/added the abstract method definitions in resource.py. The idea is that someone could implement another implementation (e.g. simulated) of a KATCP resource by basing their code on the ABC, with the ABC defining the minimum expected API that should be exposed by a KATCPResource compatible implementor.

lvdheever: I thought that is what you will say :-)
The problem was that the current implementation of set_sampling_strategy in the Container class had a bug and it set the strategy on each and every child resource. It just happened to work because the right one was amongst those. To fix it without too much ripple for now, I had to change the footprint for set_sampling_strategy in the Container class to include the resource_name. 
The sharing of the KATCPResource base class for KATCPClientResource and KATCPClientResourceContainer is somewhat awkward in some cases and causes unexpected side-effects or introduces un-tested use cases which are now started to get used (because it is available) in code which shows up bugs and anomalies between KATCPClientResource and KATCPClientResourceContainer.  
[Aargh - that just sounds unclear - better to have a chat about this some time.]

PR #124, opened 2015-09-08
logging raw katcp messages causes an error when trying to store raw messages as they contain non-ascii characters

mmphego:


katcp/client.py
```diff
@@ -504,7 +504,7 @@ def handle_message(self, msg):
         """
         # log messages received so that no one else has to
         if self._logger.isEnabledFor(logging.DEBUG):
-            self._logger.debug(str(msg))
+            self._logger.debug(repr(str(msg)))
```
theunsa: `repr(str(` intended? Why not just `repr`?

brickZA: The katcp message class does a different serialisation depending on whether
you do repr or str -- the repr one 'pretty-prints' it, which has the
disadvantage that spaces are unescaped, which means you can no longer
descern from the logs whether you have multiple arguments or a single
argument containing spaces.

theunsa: Aha ... thanks, that makes sense.

PR #126, opened 2015-09-08
Lize strategies

lvdheever:
Some cleanup to move classes and add set_sampling_strategy in addition to set_sampling_strategies


katcp/resource_client.py
```diff
@@ -735,6 +735,208 @@ def issue_request(self, *args, **kwargs):
         """
         return self._client.wrapped_request(self.name, *args, **kwargs)
```
brickZA: I'm not happy with all this being moved from resource.py to client_resource.py.  The design idea is:

resource.py specifies the abstract interface of what a KATCP resource is supposed to look like
client_resource.py specifies an implementation that uses direct KATCP connections.

The nice thing about having abstract classes is also that certain functionalities can be kept abstract in that they only depend on the implementors implementing a certain subset of the whole abstract API, see for instance the ABCs in python's builting collections module.

lvdheever: Not sure I understand clearly: 
Are you referring to the removal of the abstract method set_sensor_strategy from class KATCPResource to KATCPClientResource? In which case I can return that as and abstract method as I came up with an alternate solution that kept the footprint the same on the resource client and the container.

Or are you referring to moving the ClientGroup?  The reason was that ClientGroup operators on KATCPClientResource and not on KATCPResource directly. In which case one often had to go back to resource to align ClientGroup after changing something in resource_client. This just seemed akward and unneccessary.

PR #128, opened 2015-09-09
Lize sampling

lvdheever:
Final commit for strategy consolidation now.
Improved unit tests and tested on devll.


katcp/test/test_resource_client.py
```diff
@@ -744,34 +768,43 @@ def side_effect(*args, **kwargs):
...
-        yield DUT.set_sampling_strategy('resource1', 'sensor_1', strat1)
+        yield DUT.set_sampling_strategy('resource1_sensor_1', strat1)
         DUT.children.resource1.set_sampling_strategy.assert_called_once_with(
             'sensor_1', strat1)
         DUT.children.resource2.set_sampling_strategy.assert_not_called()
         DUT.children.resource3.set_sampling_strategy.assert_not_called()
         DUT.children.resource1.set_sampling_strategy.reset_mock()
 
-        yield DUT.set_sampling_strategy('resource2','sensor_1', strat2)
+        yield DUT.set_sampling_strategy('resource2.sensor_1', strat2)
```
theunsa: Is the dot intentional here? Or should it be underscore?

katcp/resource_client.py
```diff
@@ -1111,45 +1170,77 @@ def list_sensors(self, filter="", strategy=False, status="",
...
-            resource_obj = self.children[resource_name]
-            yield resource_obj.set_sampling_strategy(sensor_name, strategy_and_parms)
-        else:
-            # Handle aggregate sensors that are not alwasy pre-allocated to the same mon_ component
-            # TODO: Handle aggregates better
-            # (for now the aggregate sensor_obj must exist as you don't know on which resource to cache it)
-            sensor_obj = getattr(self.sensor, sensor_name, None)
-            if sensor_obj:
-                resource_obj = self.children[sensor_obj.parent_name]
+    def _resource_set_sampling_strategies(self, resource_name, sensor_name, strategy_and_parms):
```
theunsa: Just check max-line-length throughout.

katcp/test/test_resource_client.py
lvdheever: Yes intentional - I wanted to do one with a dot and one with an underscore to check that both work.

katcp/resource_client.py
lvdheever: Ok - always forget that one as it is superficial to me. But then, rules are rules.

PR #129, opened 2015-09-09
Adding timeout functionality for all katcp 'until_' methods

theunsa:


lvdheever:
Loos good.
I am just wondering if a default timeout will be better than None?
I know None is mapping to current behaviour, but may be an actual timeout will be better?
But now, typing this up, I have second thoughts and think the timeout of None is what one would expect from an "until_xxx" method, unless otherwise specified.
Will merge this.

katcp/client.py
```diff
@@ -923,11 +923,17 @@ def is_connected(self):
         return self._connected.isSet()
 
     @tornado.gen.coroutine
-    def until_connected(self):
+    def until_connected(self, timeout=None):
         """Return future that resolves when the client is connected."""
-        yield self.until_running()
+        t0 = self.ioloop.time()
+        yield self.until_running(timeout=timeout)
+        t1 = self.ioloop.time()
```
brickZA: This is a case where utils.timeoutmanager would be your friend I think?

PR #134, opened 2015-11-02
Use request timeout in ioloop wrapper

ludwigschwardt:
The timeout parameter on the decorator itself is never used. Rather get rid
of it and extract the relevant timeout from the kwargs of the original request.
This way the user can override timeouts per request (as opposed to per ioloop
wrapper).

A future improvement might be to make the ioloop timeout slightly longer than
the request timeout to allow the request to drive the error reporting to the
user.

@brickZA to have a look as well.

### DIFF
--- a/katcp/resource_client.py
+++ b/katcp/resource_client.py
@@ -1380,10 +1380,12 @@ class IOLoopThreadWrapper(object):
             # Should never get here since the tornado future should raise
             assert False, 'Tornado Future should have raised'
 
-    def decorate_callable(self, callable_, timeout=None):
+    def decorate_callable(self, callable_):
         """Decorate a callable to use call_in_ioloop"""
         @wraps(callable_)
         def decorated(*args, **kwargs):
+            # Extract timeout from request itself or use default for ioloop wrapper
+            timeout = kwargs.get('timeout')
             return self.call_in_ioloop(callable_, args, kwargs, timeout)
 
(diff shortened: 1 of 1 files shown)


katcp/resource_client.py
```diff
@@ -1380,10 +1380,12 @@ def call_in_ioloop(self, fn, args, kwargs, timeout=None):
             # Should never get here since the tornado future should raise
             assert False, 'Tornado Future should have raised'
 
-    def decorate_callable(self, callable_, timeout=None):
+    def decorate_callable(self, callable_):
         """Decorate a callable to use call_in_ioloop"""
         @wraps(callable_)
         def decorated(*args, **kwargs):
+            # Extract timeout from request itself or use default for ioloop wrapper
+            timeout = kwargs.get('timeout')
```
theunsa: `get` will return empty string as default which might not be interpreted correctly in the callable ... would suggest explicit `None` as default when timeout argument not present.

ludwigschwardt: Not true... If kwargs is a standard dict it returns None by default.

dict.get?
D.get(k[,d]) -> D[k] if k in D, else d.  d defaults to None.

theunsa: Sorry, my bad ... I screwed up when I tested this.

theunsa:
Happy, to merge ... will hand over to Neilen

PR #136, opened 2016-01-28
Improve latency in katcp client loops CB-568

brickZA:
If you receive a future that is already done (i.e. fut.done() == True) before yielding on itin a tornado coroutine, the yield statement will not result in a pass being made through the ioloop, effectively turning your coroutine into a blocking function. 

By timing for how long only 'done' futures are received one can adaptively force a trip through the ioloop by doing yield tornado.gen.moment when things have been blocking for too long.


theunsa:
Wow ... nice and hardcore @brickZA

PR #139, opened 2016-02-01
Suppress spurious logs when multiple futures that are being waited on time out

brickZA:


martinslabber:
Nice use of gen.multi and quiet_exceptions, its new to me.

PR #141, opened 2016-04-20
Add a client-side dummy-request facility CB-918

brickZA:
@martinslabber @profcalculus Please have a look.

This implements the neccesary client side functionality to allow 'dummy-requests' to be done.  Next move would be modifying katsubarray to add the 'dummy_unknown_requests' flag to the resource client configuration dicts as appropriate.

### DIFF
--- a/katcp/core.py
+++ b/katcp/core.py
@@ -14,7 +14,7 @@ import logging
 
 import tornado
 
-from collections import namedtuple
+from collections import namedtuple, defaultdict
 from functools import wraps, partial
 
 from concurrent.futures import Future, TimeoutError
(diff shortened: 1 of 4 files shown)


lvdheever:
I presume you guys had a good chat about this approach and why it is being handled in katcp. To me, at first glance without having been involved in any of the discussions and without thinking it through it feels harsh to add this support in the katcp library. Somehow it feels like it should be handled higher up, somewhere in the container or such. But there are probably problems if doing it higher up in the hierarchy. At least it is in resource.py

brickZA:
@lvdheever It turns out it is not trivial to handle at a single level, since the proxy cannot know what requests it needs to fake if it is not connected to the devices (spmc / cbf) in question. The basic idea is that 1) subarray manager will instruct the data proxy client to fake missing requests (using above implementation), subaray manager will instruct the proxy to hide all data related requests (triggering the client behaviour) as in https://github.com/ska-sa/katproxy/pull/193.

Apart from that, Ludwig and I have had the idea of adding a client-side fake mode for dry running and other testing needs for a while in any case, so this functionality should find other uses going forward.

PR #146, opened 2016-06-07
Add dummy unknown requests to ClientGroups

ludwigschwardt:
ClientGroups have req objects and therefore also qualify for dummy
unknown requests. Since a group is not constructed from resource specs,
provide dummy requests if any of its underlying clients have them.


ludwigschwardt:
Let's try again @brickZA :-)

PR #147, opened 2016-06-14
Documentation and minor code updates for release CB-1150

brickZA:


martinslabber:
Thanks, for the updated docs and especially for the examples.

ludwigschwardt:
Will this be rolled out to site too? Would be nice to fix this SB
verification business :-)

martinslabber:
Site is about two weeks behind at the moment. Lab was upgraded yesterday. 
On the next big upgrade on site we will update all packages. Not scheduled but we hope soon.

PR #154, opened 2016-06-27
Charles testutils fix

profcalculus:


profcalculus:
katcore #39 and katproxy #218 both depend on this PR.

katcp/server.py
```diff
@@ -447,6 +447,7 @@ def _install(self):
         # Do stuff to put us on the IOLoop
         self.ioloop_thread_id = get_thread_ident()
         self._tcp_server.add_socket(self._server_sock)
+        threading.currentThread().name = getattr(self, 'name', self.__class__.__name__)
```
brickZA: Multiple servers sharing an ioloop might cause them to interfere, but no real harm I guess.

PR #157, opened 2016-07-27
Allow high level clients to preset KATCP flags, allows broken servers

brickZA:
To help with CBF tests


katcp/test/test_resource_client.py
```diff
@@ -804,6 +804,28 @@ def handler(self, req, msg):
             s._request_handlers['sparkling-new-'+s_name] = handler
 
     @tornado.testing.gen_test
+    def test_preset_protocol_flags(self):
+        pf1 = ProtocolFlags(4,0,'')
```
martinslabber: Trivial: But you would like to know.
Missing spaces after commas. Lines 808,809 and 810

PR #158, opened 2016-07-29
Clean up ClientGroup wait()

ludwigschwardt:
Multiple TimeoutErrors are potentially raised when waiting on multiple futures
with their own (equal) timeouts, leading to messy error messages. Clean this
up by using tornado.gen.multi with the appropriate exception quieting.

Since this trick is already in use in the module, put it in a separate
helper function. The name is chosen to indicate kinship with 'until_any'
although the API is slightly different (i.e. timeouts are delegated to
individual futures).

Also improve the docs around timeouts.

### DIFF
--- a/katcp/resource.py
+++ b/katcp/resource.py
@@ -197,18 +197,21 @@ class KATCPResource(object):
             Wait for this status, at the same time as value above, to be
             obtained. Ignore status if None
         timeout : float or None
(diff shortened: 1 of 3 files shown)


ludwigschwardt:
I've decided to abandon this in favour of a TimeoutError-less wait... Watch this space :-)

PR #159, opened 2016-07-29
Simplify sensor wait() interface

ludwigschwardt:
The current wait() machinery is unsatisfactory. If the sensor
meets the condition, wait() returns True. On the other hand,
if the sensor does not meet the condition before the timeout,
a TimeoutError is raised. Wait() never returns False, making
the whole thing quite asymmetric.

The TimeoutErrors currently cause observation scripts to
crash (mostly for no good reason). Catching the exception
within the observation framework is messy.

My proposal is to swallow the TimeoutError at the KATCPResource
level and turn it into a proper symmetric False return.
User-facing client and client group objects will therefore
never crash like this again.

In addition, the ClientGroup.wait now returns extra information
that allows the user to report back on which clients have failed
the sensor condition without re-evaluating those conditions.
Users who don't care about the extra info can still simply check
the truth value of the returned dict.

As a further minor niggle, the explicit status check is dropped
in favour of the existing sensor reading check that can handle
all cases of interest.

Since the session and observation scripts are the main users of
wait(), I don't foresee any issues with the API change. All scripts
in katsdpscripts master remain valid.

This addresses JIRA tickets [MKAIV-244](https://skaafrica.atlassian.net/browse/MKAIV-244), [MRTS-361](https://skaafrica.atlassian.net/browse/MRTS-361) and [MKAIV-126](https://skaafrica.atlassian.net/browse/MKAIV-126) and related ones (and also [KAT-1869](https://skaafrica.atlassian.net/browse/KAT-1869) to some extent).

### DIFF
--- a/katcp/resource.py
+++ b/katcp/resource.py
@@ -857,7 +857,7 @@ class KATCPSensor(object):
             condition is satisfied. Since the reading is passed in, the value,
             status, timestamp or received_timestamp attributes can all be used
             in the check.
-            TODO: Sequences of conditions (use SensorTranstionWaiter thingum?)
+            TODO: Sequences of conditions (use SensorTransitionWaiter thingum?)
         timeout : float or None
             The timeout in seconds (None means wait forever)
 
--- a/katcp/resource.py
+++ b/katcp/resource.py
@@ -200,16 +200,14 @@ class KATCPResource(object):
         Returns
         -------
         This command returns a tornado Future that resolves with True when the
-        sensor value satisfies the condition. It will never resolve with False;
-        if a timeout is given a TimeoutError happens instead.
+        sensor value satisfies the condition, or False if the condition is
+        still not satisfied after a given timeout period.
 
         Raises
         ------
         :class:`KATCPSensorError`
             If the sensor does not have a strategy set, or if the named sensor
             is not present
-        :class:`tornado.gen.TimeoutError`
-            If the sensor condition still fails after a stated timeout period
 
         """
         sensor_name = escape_name(sensor_name)
--- a/katcp/resource.py
+++ b/katcp/resource.py
@@ -180,6 +180,7 @@ class KATCPResource(object):
     def children(self):
         """AttrDict of subordinate KATCPResource objects keyed by their names."""
 
+    @tornado.gen.coroutine
     def wait(self, sensor_name, condition_or_value, status=None, timeout=5):
         """Wait for a sensor in this resource to satisfy a condition.
 
@@ -216,7 +217,12 @@ class KATCPResource(object):
         """
         sensor_name = escape_name(sensor_name)
         sensor = self.sensor[sensor_name]
-        return sensor.wait(condition_or_value, status, timeout)
+        try:
+            yield sensor.wait(condition_or_value, status, timeout)
+        except tornado.gen.TimeoutError:
+            raise tornado.gen.Return(False)
+        else:
+            raise tornado.gen.Return(True)
 
     @abc.abstractmethod
     def list_sensors(self, filter="", strategy=False, status="",
--- a/katcp/resource.py
+++ b/katcp/resource.py
@@ -181,7 +181,7 @@ class KATCPResource(object):
         """AttrDict of subordinate KATCPResource objects keyed by their names."""
 
     @tornado.gen.coroutine
-    def wait(self, sensor_name, condition_or_value, status=None, timeout=5):
+    def wait(self, sensor_name, condition_or_value, timeout=5):
         """Wait for a sensor in this resource to satisfy a condition.
 
         Parameters
@@ -194,9 +194,6 @@ class KATCPResource(object):
             condition is satisfied. Since the reading is passed in, the value,
             status, timestamp or received_timestamp attributes can all be used
             in the check.
-        status : int enum, key of katcp.Sensor.SENSOR_TYPES or None
-            Wait for this status, at the same time as value above, to be
-            obtained. Ignore status if None
         timeout : float or None
             The timeout in seconds (None means wait forever)
 
@@ -218,7 +215,7 @@ class KATCPResource(object):
         sensor_name = escape_name(sensor_name)
         sensor = self.sensor[sensor_name]
         try:
-            yield sensor.wait(condition_or_value, status, timeout)
+            yield sensor.wait(condition_or_value, timeout)
         except tornado.gen.TimeoutError:
             raise tornado.gen.Return(False)
         else:
@@ -851,7 +848,7 @@ class KATCPSensor(object):
         # By now the sensor manager should have set the reading
         raise Return(self._reading.status)
 
-    def wait(self, condition_or_value, status=None, timeout=None):
+    def wait(self, condition_or_value, timeout=None):
         """Wait for the sensor to satisfy a condition.
 
         Parameters
@@ -863,9 +860,6 @@ class KATCPSensor(object):
             status, timestamp or received_timestamp attributes can all be used
             in the check.
             TODO: Sequences of conditions (use SensorTranstionWaiter thingum?)
-        status : int enum, key of katcp.Sensor.SENSOR_TYPES or None
-            Wait for this status, at the same time as value above, to be
-            obtained. Ignore status if None
         timeout : float or None
             The timeout in seconds (None means wait forever)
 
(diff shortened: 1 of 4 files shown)


katcp/resource.py
```diff
@@ -853,30 +857,24 @@ def wait(self, condition_or_value, status=None, timeout=None):
             condition is satisfied. Since the reading is passed in, the value,
             status, timestamp or received_timestamp attributes can all be used
             in the check.
-
-            Sequences of conditions are TODO, use SensorTranstionWaiter thingum?
-        status : int enum, key of katcp.Sensor.SENSOR_TYPES or None
-            Wait for this status, at the same time as value above, to be
-            obtained. Ignore status if None
+            TODO: Sequences of conditions (use SensorTranstionWaiter thingum?)
```
martinslabber: s/SensorTranstionWaiter/SensorTransitionWaiter/

katcp/resource.py
```diff
@@ -894,9 +892,7 @@ def handle_update(sensor, reading):
             # This handler is called whenever a sensor update is received
             try:
                 assert sensor is self
-                cond_matched = condition_test(reading)
-                status_matched = reading.status == status or status is None
```
theunsa: This explicit status check was originally included by @brickZA ... would like to hear his thoughts about dropping it.

lvdheever: Agree - @brickZA  should give the go-ahead for this merge

katcp/resource_client.py
```diff
@@ -55,6 +55,23 @@ def _transform(f):
...
 
+def until_all(futures):
+    """Run multiple operations in parallel but only raise a single TimeoutError.
+
+    This returns a yieldable object that will contain the results of `futures`
+    in an equivalent container (list or dict). However, it only raises the
+    first TimeoutError and ignores any further timeouts. This makes it useful
+    to wait on the results of multiple operations with the same timeout, as the
+    first occurrence of a timeout means that any unresolved futures will also
+    necessarily time out, leading to a rash of duplicate exceptions otherwise.
```
theunsa: > as the first occurrence of a timeout means that any unresolved futures will also necessarily time out

Is bit of an assumption ... but ok I guess, as long as the other future calls will get a chance to finish and not be interrupted by the first future call with a TimeoutError - its not clear to me from the documentation if this is the case though:
> If any children raise exceptions, `multi(`) will raise the first one. All others will be logged, unless they are of types contained in the `quiet_exceptions` argument.

katcp/resource_client.py
```diff
@@ -55,6 +55,23 @@ def _transform(f):
...
+
+    The function name implies kinship with :func:`katcp.core.until_any`.
+    The primary difference is that `until_any` takes an explicit timeout while
+    this function relies on the individual futures to have their own timeouts.
+
```
theunsa: Aha ... the concern in my previous comment is rather with `katcp.core.until_any`. So the "primary difference" for me between `until_any` and `until_all` is rather that the former returns when **any** of the provided futures resolves.

theunsa:
@ludwigschwardt: 

Looks like a good change to me. I would like @brickZA to do the final check and merge given that he is the original author of most of this.

CAM also makes use of the `wait`s although more indirectly through `until_synced` etc but I don't see any impact on our code due to this change ... @brickZA can just confirm the removal of the sensor status part in the condition check.

katcp/resource.py
ludwigschwardt: Not true on both counts...

`condition_test` takes an entire reading and is free to check value, status and/or timestamps. You are maybe confused with the case where `condition_or_value` is an actual value as opposed to a generic function :-)

The explicit status check was introduced by yours truly in the katuilib days already. The evidence:

> r12268 | schwardt | 2012-01-24 09:59:41 +0200 (Tue, 24 Jan 2012) | 1 line
> 
> (schwardt, simonr) Added the ability to wait until an arbitrary condition is satisfied on a sensor, in addition to the default condition of the sensor value being equal to a desired value. The differing wait implementations in KATClient and Array are now unified with the same parameters and behaviour. The signal implementation in KATClient has been replaced by a time-based implementation, which is shorter and slightly easier to understand. The timeouts were made more accurate. Both versions now throw exceptions on missing sensors or ones without a strategy, and both respond to a KeyboardInterrupt. The updated signal implementation in KATClient is kept in commented-out form for posterity.
> 

That is why one writes decent commit messages... Unfortunately it was not decent enough to remind me of the need for an explicit status check :-)

I think my thinking was that you always only want to wait on a 'nominal' sensor, and if that is the case you might as well make it explicit instead of forcing a lambda-type check all the time.

Then, in the translation to katcp, the default became `status=None`, so that whole thing went out the door. But nothing much went wrong, so I guess my need for 'nominal'-ness has been misplaced :-) It sort of makes sense, as warn and error states do not invalidate the actual value of the sensor and should still be wait()-able.

katcp/resource_client.py
ludwigschwardt: Sure. I should tone down that word 'primary'. I assumed the names were suggestive enough of the meaning of the functions but wanted to point out a difference in their APIs. `until_any` takes an explicit `timeout` parameter while `until_all` assumes that the futures are already wrapped in timeouts themselves.

ludwigschwardt:
I discussed all this with @brickZA before I started, so I'm hoping for a blessing :-)

As pointed out, the `wait()` machinery really dates back to the early days and Neilen kindly dragged it kicking and screaming into the async world.

CAM makes no use of `wait` as far as I can tell. The `until_synced` method and friends relies on the `AsyncState` class which is a completely separate waiting framework.

lvdheever:
Wonder what happened to @brickZA - he's very very quite - although I hear him typing furiously at this desk :-)

katcp/resource.py
theunsa: Yes `condition_test` does take entire reading and is free to check whatever but my point was that it does not:
```
882         condition_test = (condition_or_value if callable(condition_or_value)
                              else lambda s: s.value == condition_or_value)
```

But since you are the original author and happy with removing the status check, it does not really matter. Your reasoning makes sense to me.

theunsa:
@ludwigschwardt: Your right ... I had a closer look and also cannot find `wait` use in CAM code.

katcp/resource_client.py
```diff
@@ -55,6 +55,23 @@ def _transform(f):
     future.add_done_callback(_transform)
     return new_future
 
+def until_all(futures):
+    """Run multiple operations in parallel but only raise a single TimeoutError.
+
```
brickZA: I'm fine with everything, except until_all seems a bit uneccesary to me. until_any was only introduced since tornado did not have  a similar function. Arguably it has been superceded by more recent tornado functionality. 

Calling it until_all is confusing IMHO, since that is also what I would call the default tornado yield [f1, f2, f...] behaviour, with slight differences in multiple exception handling. 

I would recommend just using  tornado.gen.multi directly. It is well documented by tornado, etc.

katcp/resource.py
ludwigschwardt: Let me explain myself better here... The `condition_or_value` parameter passed to `wait` is either a callable taking a sensor reading as parameter, or the desired sensor value itself. Line 882 quoted above simply normalises both cases to be a function for ease of use.

In the first case the parameter is already a function so becomes `condition_test` directly (and can check the status if it wants to). In the second case, we construct a function that checks the sensor value against the desired value in `condition_or_value`. Did I miss something?

katcp/resource_client.py
ludwigschwardt: Agreed, will make this fix, as I do not care about `until_all` :-)

katcp/test/test_resource.py
```diff
@@ -94,8 +94,10 @@ def test_wait(self):
         # wait for value and status
         waiting_value = 3
         waiting_status = 'warn'
+        waiting_condition = lambda reading: reading.value == waiting_value and \
+                                            reading.status == waiting_status
         self.io_loop.add_callback(DUT.set_value, 3, Sensor.WARN)
-        yield DUT.wait(waiting_value, waiting_status)
+        yield DUT.wait(waiting_condition, timeout=timeout)
```
ludwigschwardt: @theunsa, see also this test to set your mind at ease that status checking still works :-)

katcp/resource.py
theunsa: Agree, I wanted to make sure we did not want to add status checking in the non-callable case as well. Given that we agree that the satus-checking is not necessary its superflous to discuss this anymore.

ludwigschwardt: Cool, thanks for the clarification!

ludwigschwardt:
OK, `until_all` is gone again...

PR #161, opened 2016-08-18
CB-1320 - plug memory leak in KATCPClientResourceContainer

fjoubert:
removed __del__ method that calls the join on katcp_client (which in turn calls join on the io_loop_manager's thread join) because it was causing a massive memory leak for anyone using the KATCPClientResourceContainer

### DIFF
--- a/katcp/inspecting_client.py
+++ b/katcp/inspecting_client.py
@@ -164,9 +164,6 @@ class InspectingClientAsync(object):
                 connected=False, synced=False, model_changed=False, data_synced=False)
         )
 
-    def __del__(self):
-        self.close()
-
     def inform_hook_client_factory(self, host, port, *args, **kwargs):
         """Return an instance of :class:`_InformHookDeviceClient` or similar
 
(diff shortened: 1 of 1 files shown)


katcp/inspecting_client.py
```diff
@@ -353,14 +350,14 @@ def set_state_callback(self, cb):
...
         self.stop()
-        self.join()
 
     def start(self, timeout=None):
         return self.connect(timeout)
 
     def stop(self, timeout=None):
         self._running = False
         self.katcp_client.stop(timeout)
+        self.join(timeout)
```
brickZA: Stop should not join, just remove this line.

katcp/inspecting_client.py
```diff
@@ -164,9 +164,6 @@ def __init__(self, host, port, ioloop=None, initial_inspection=None,
                 connected=False, synced=False, model_changed=False, data_synced=False)
         )
 
-    def __del__(self):
```
brickZA: Perhaps document that stop/close needs to be called to ensure cleanup in the class docstring or perhaps in start's docstring?

fjoubert: Added some docstrings on the class and on the start method in 39680a9

katcp/inspecting_client.py
fjoubert: Removed in 39680a9

PR #163, opened 2016-08-26
remove legacy version str creation and use katversion

fjoubert:


ludwigschwardt:
Awesome!

There is also katcp/version.py that could probably be removed in its entirety (used in very few places).

PR #164, opened 2016-08-26
Release notes for version 0.6.1; removed version.py because we use katversion now

fjoubert:
...


fjoubert:
When this is merged I will tag HEAD with v0.6.1 and publish to pypi.

ludwigschwardt:
You'll need to fix `__init__.py` too. It still imports version.py. There is also an `except ImportError` related to version.py that can go if you ask me.

See also `doc/conf.py`.

fjoubert:
Good catch - include `__init__.py` changes in 18cc413.

ludwigschwardt:
Mmm,

I see more issues...

`katcp/server.py` also reports the KATCP version via version.py (see also VERSION and VERSION_STR)

And there is the Sphinx docs script in `doc/conf.py`.

fjoubert:
I am in the process of reworking katcp/server.py to be better with the whole version thing. This PR is not ready to be merged yet.

Thanks for the inputs Ludwig!

fjoubert:
Please review with latest changes included! (9c24471 and 2cb8bc6)

katcp/server.py
```diff
@@ -34,9 +34,11 @@
 from .sampling import format_inform_v5, format_inform_v4
 from .core import (SEC_TO_MS_FAC, MS_TO_SEC_FAC, SEC_TS_KATCP_MAJOR,
                    VERSION_CONNECT_KATCP_MAJOR, DEFAULT_KATCP_MAJOR)
-from .version import VERSION, VERSION_STR
 from .kattypes import (request, return_reply)
 
+from katversion import get_version
```
ludwigschwardt: The `get_version` function is used to extract versions from git repos, but the git repo won't be around once you install katcp.

I would rather do `from . import __version__` or something similar. You should restrict `katversion` to `setup.py` and `__init__.py` and use the explicit `__version__` elsewhere.

katcp/test/test_server.py
```diff
@@ -28,6 +28,7 @@
     start_thread_with_cleanup, WaitingMock, ClientConnectionTest, mock_req,
     handle_mock_req)
 from katcp.core import FailReply
+import katversion
```
ludwigschwardt: Ditto.

doc/conf.py
```diff
@@ -19,7 +19,7 @@
 sys.path.insert(0, os.path.abspath('..'))
 sys.path.insert(0, os.path.abspath('./exts'))
 
-import katcp
+import katversion
```
ludwigschwardt: Ditto (see comment in server.py).

katcp/server.py
fjoubert: Ah that makes sense, I was a bit unsure of how to handle it, this explains it nicely. I'm a bit of a noob when it comes to packaging + versioning, but I'm learning 👍

katcp/server.py
```diff
@@ -34,9 +34,12 @@
 from .sampling import format_inform_v5, format_inform_v4
 from .core import (SEC_TO_MS_FAC, MS_TO_SEC_FAC, SEC_TS_KATCP_MAJOR,
                    VERSION_CONNECT_KATCP_MAJOR, DEFAULT_KATCP_MAJOR)
-from .version import VERSION, VERSION_STR
 from .kattypes import (request, return_reply)
 
+# 'import katcp' so that we can use katcp.__version__ later
+# we cannot do this: 'from . import __version__' because __version__
+# does not exist at this stage
+import katcp
```
fjoubert: I am not sure why this is needed or why __version__ does not exist at this point in time, but I cannot import __version__ in server.py from katcp because it does not exist yet. Is this good enough, or do you know of a better way?

ludwigschwardt: I think `__init__.py` needs to run in order to generate `__version__` and I'm not sure why the relative import does not trigger this...

fjoubert:
Any more comments? @brickZA have you had a look?

katcp/server.py
ludwigschwardt: Just to clarify this: `server.py` is imported within `\__init\__.py` before `\__version\__` exists.

The one way around this circular import is to split out `\__version\__` to its own file, i.e. restore `version.py` which we just removed. :-)

The other way is to `import katcp` in full as is done here. A bit brute force but does the job. Revisit if needed.

ludwigschwardt:
I am happy if @brickZA is happy with the katcp import mentioned.

CHANGELOG
```diff
@@ -1,3 +1,16 @@
+ (0.6.1)
+ * Increased the maximum memory size that can be received up from 128k to 2M
```
brickZA: Should that be "maxiumum _message_ size" ?

katcp/server.py
brickZA: Would it not work to change the order of statements in __init__.py such that __version__ is generated before it imports from other submodules? Otherwise we have a bit of a circular import which can become fun IIRC. In any case, I don't have strong feelings about this if what you did works already, just a comment.

ludwigschwardt: The problem is that katversion insists on putting `__version__` at the end of `__init__.py` to ensure that it is authoritative. I'm not sure what will happen if you put say `__version__ = None` at the top and then replace it at the end (will `server.py` see the change?).

A separate `version.py` is the cleanest but again requires changes to katversion. I prefer the current iteration of katversion as it is simpler for most packages to just stick to `__init__.py`.

The current solution was also touted [here](http://stackoverflow.com/questions/1556387/circular-import-dependency-in-python).

PR #167, opened 2016-10-18
Make parsing of int's safer, can now handle a float string.

martinslabber:
We see that at times floating point strings are given to the Int sensor. To make the decoding more robust, cast it as float then as string.

### DIFF
--- a/katcp/kattypes.py
+++ b/katcp/kattypes.py
@@ -158,7 +158,7 @@ class Int(KatcpType):
 
     def decode(self, value, major):
         try:
-            return int(value)
+            return int(float(value))
         except:
             raise ValueError("Could not parse value '%s' as integer." % value)
 
@@ -293,7 +293,7 @@ class Lru(KatcpType):
     # LRU sensor values
     LRU_NOMINAL, LRU_ERROR = range(2)
 
-    ## @brief Mapping from LRU value constant to LRU value name.
+    # @brief Mapping from LRU value constant to LRU value name.
     LRU_VALUES = {
         LRU_NOMINAL: "nominal",
         LRU_ERROR: "error",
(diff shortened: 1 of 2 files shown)


bmerry:
Could there be situations where this causes an int to lose precision, by first casting it to float then back to int? 48-bit timestamps should round-trip via float, but are there any 64-bit int sensors?

It might be safer to only go via float if the direct conversion throws `ValueError`.

martinslabber:
Good point @bmerry.
Agree it would be better to try an alternative conversion only once a problem occurred. In that case I'll cast it as Decimal rather than float. I'll update the PR a bit later today.

Thinking aloud:
``` 
from decimal import Decimal
...
try:
    try:
        return int(value)
    except ValueError:
        return int(Decimal(value))
except:
    raise ValueError("Could not parse value '%s' as integer." % value)
```

brickZA:
Hold on, don't we _want_ to raise an error if a float is passed to an int parameter?

martinslabber:
Raising when you get "0.0" on Integer sensor type is debatable, obviously I think its better to be robust and carry on thus the PR. Float type work if you pass it "0", but there is a StrictFloat type for when you demand it, if I read the code correctly.

I made the change when there was an problem between CAM and CBF and CBF was sending "0.0" to an Integer sensor, CBF has been fixed but we all though that KATCP should be a bit more robust. 
I'll leave to @brickZA to decide if this should go in or not.

bmerry:
I'm also pretty uncomfortable with this - if someone has mis-declared a sensor as Integer when the values are float, we should fail early, rather than silently truncate the value and then discover months later that all the stored sensor values are useless.

brickZA:
I'm with @bmerry on this one, so unless there are any strong counter arguments I will close the PR?

martinslabber:
It was tabled and decided not to do this.

PR #168, opened 2016-10-20
Properly handle errors during inspection in inspecting client

brickZA:
On the dev32 system the single process that is trying to connect and sync to all 32
dishes takes too long, resulting in some of the inspection requests timing
out. InspectingClient was ignoring these errors, resulting in a client that says
it is synced, but does not properly reflect its server's state

Also added a random-exponential retry backoff process to ensure that the
thundering herd does not continue thundering

### DIFF
--- a/katcp/inspecting_client.py
+++ b/katcp/inspecting_client.py
@@ -52,12 +52,18 @@ class ExponentialRandomBackoff(object):
         self.randomicity = randomicity
         self._update_delay()
 
-    def _update_delay(self):
-        r = self.randomicity
-        assert 0 <= r <= 1
+    def _validate_parameters(self):
+        # Since a user could potentially change these parameters at any time and
+        # since it is nice to give feedback at __init__ time as well, have a
+        # single validation routine that can be called any place it matters
         assert 0 <= self.delay_initial <= self.delay_max
-        bd = self._base_delay
-        d = (1 - r)*bd + r*random.random()*bd
+        assert 0 <= self.randomicity <= 1
+        assert self.exp_fac > 1
+
+    def _update_delay(self):
+        self._validate_parameters()
+        r = self.randomicity
+        d = ((1 - r) + r*random.random())*self._base_delay
         self._delay = min(d, self.delay_max)
 
     def failed(self):
@@ -66,10 +72,10 @@ class ExponentialRandomBackoff(object):
         After calling failed(), the `delay` property contains the next delay
         """
         try:
-            bd = self._base_delay
-            exp_fac = self.exp_fac
-            assert exp_fac > 1
-            self._base_delay = min(bd * exp_fac, self.delay_max)
+            self._validate_parameters()
+            self._base_delay = min(
+                self._base_delay * self.exp_fac,
+                self.delay_max)
             self._update_delay()
         except Exception:
             ic_logger.exception(
@@ -77,7 +83,7 @@ class ExponentialRandomBackoff(object):
 
     def success(self):
         """Call whenever an action has succeeded, resets delay to minimum"""
-        self._base_delay = self.delay_min
+        self._base_delay = self.delay_initial
         self._update_delay()
 
     @property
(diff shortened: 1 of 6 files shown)


katcp/inspecting_client.py
```diff
@@ -109,12 +174,15 @@ class InspectingClientAsync(object):
     """
 
     sync_timeout = 5
```
lvdheever: I also tested this on dev32 and it is working well. 

However, my suggestion is it increase the sync_timeout to 10s. It is a timeout after all and if it completes before the timeout triggers then all is OK. If it actually still was busy trying to complete the syncing, then backing off and retrying is just adding to the load. 
The sensor-list timeouts reported were all between 1.5 to 3.8 seconds. I tested on dev32 with 8s timeout and it all completed successfully on first time sync then. I suggest to set the sync_timeout to 10s though. That way we have a good chance of syncing up on the first try, and does not hit the lower level device too hard by resyncing often and we still have the retry mechanism if that does not work.

katcp/inspecting_client.py
```diff
@@ -21,6 +22,70 @@
...
+            Fraction of timeout that should be randomly calculated. If
+            randomicity is 0, the base delay will always be used, if it is 1, a
+            random value between 0 and the base delay is calculated, or the
+            weighted average of the two for intermediate values.
+
+        """
+        self.delay_initial = delay_initial
+        self.delay_max = delay_max
+        self.exp_fac = exp_fac
+        """Increase timeout by this factor for each consecutive failure"""
```
brickZA: This is actually the proper way to document an attribute in Python, sphinx will pick it up, etc.

IIRC the rules are that if a string that follows the definition of an attribute in the class body or in `__init__` will be taken as the docstring for that attribute.

katcp/inspecting_client.py
```diff
@@ -21,6 +22,70 @@
...
+        self._base_delay = self.delay_initial
+        self.randomicity = randomicity
+        self._update_delay()
+
+    def _update_delay(self):
+        r = self.randomicity
+        assert 0 <= r <= 1
+        assert 0 <= self.delay_initial <= self.delay_max
```
brickZA: I am going under the conservative assumption (since it is a public member) that a user could change the value at any time. Thoughts?

katcp/inspecting_client.py
```diff
@@ -109,12 +174,15 @@ class InspectingClientAsync(object):
...
+    initial_resync_timeout = 1
+    max_resync_timeout = 90
 
     def __init__(self, host, port, ioloop=None, initial_inspection=None,
                  auto_reconnect=True, logger=ic_logger):
-        # TODO Consider optinal 'name' parameter just to make logging clearer
+        # TODO Consider optional 'name' parameter just to make logging clearer
         self._logger = logger
+        self.resync_delay = None
+        """Set to an ExponentialRandomBackoff instance in _state_loop"""
```
brickZA: Ditto :)

katcp/inspecting_client.py
brickZA: As discussed IRL the proper solution should be addressed in CB-1609. I think we should not go down the route of arbitrarily long timeouts since that tends to mask real problems while making a system lazy in the face of trouble :) 

As a compromise increased to 8s with a long TODO comment :)

katcp/test/test_inspectclient.py
```diff
@@ -363,10 +389,11 @@ def test_from_connect(self):
         self.assertEqual(state, inspecting_client.InspectingClientStateType(
             connected=True, synced=False, model_changed=False, data_synced=False))
         self.assertIs(model_changes, None)
-        # Due to the structure of the state loop the initial state may be sent twice
-        # before we get her, and was the case for the initial implementation. Changes made
-        # on 2015-01-26 caused it to happy only once, hence + 1. If the implementation
-        # changes having + 2 would also be OK.
+        # Due to the structure of the state loop the initial state may be sent
+        # twice before we get here, and was the case for the initial
+        # implementation. Changes made on 2015-01-26 caused it to happy only
```
brickZA: Fixed

katcp/inspecting_client.py
brickZA: I've refactored the validation, source code comment should hopefully explain the idea.

brickZA:
OK, think I have addressed all comments.

lvdheever:
@ajoubertza - please review ASAP. Tx

PR #170, opened 2016-10-28
CB-1591 fix partial sync failure case

brickZA:
When sensor sync failed, requests model updates were never sent by inspecting
client callback

### DIFF
--- a/katcp/inspecting_client.py
+++ b/katcp/inspecting_client.py
@@ -7,6 +7,7 @@ from __future__ import print_function
 
 import logging
 import random
+import copy
 
 import tornado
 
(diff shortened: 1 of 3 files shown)


katcp/inspecting_client.py
```diff
@@ -322,6 +331,15 @@ def until_data_synced(self, timeout=None):
...
 
+    def until_state(self, desired_state, timeout=None):
+        """
+        Wait until state is desired_state, InspectingClientStateType instance
+
+        Returns a future
+
+        """
+        return self._state.until_state(desired_state)
+
```
lvdheever: Did you not want to pass the timeout on?

PR #175, opened 2016-12-13
Use instance logger in resource_client CB-1731

brickZA:


katcp/resource_client.py
```diff
@@ -554,53 +554,53 @@ def _request_factory(self, name, description):
...
-                log.debug('{}: Setting state to "syncing"'.format(self.address_string))
+                self._logger.debug('{}: Setting state to "syncing"'.format(self.address_string))
                 self._state.set_state('syncing')
                 if model_changes:
-                    log.debug('{}: handling model updates: {}'.format(
+                    self._logger.debug('{}: handling model updates: {}'.format(
                         self.address_string,model_changes))
                     yield self._update_model(model_changes)
-                    log.debug('{}: finished handling model updates'
+                    self._logger.debug('{}: finished handling model updates'
```
martinslabber: FYI: The indentation of lines 558, 574, 577 and a few more are now out. 
We should really consider, at some point, improving the debug logging in katcp by using the more clunky args style and save a few clock cycles with all these debug messages.

PR #177, opened 2017-01-11
First stab at concurrent async request handlers CB-1747

brickZA:


brickZA:
@bmerry Still need to add tests and docs, but see if this works for you.

katcp/server.py
```diff
@@ -1071,16 +1075,21 @@ def async_reply(f):
...
-                    # a thread-safe future. Supporting sync-with-thread and
-                    # async futures is turning out to be a pain in the ass ;)
-                    self.ioloop.add_callback(reply.add_done_callback, async_reply)
-                    # reply.add_done_callback(async_reply)
-                    self._logger.debug("%s FUTURE OK" % (msg.name,))
-                    return done_future
+                    if getattr(handler, 'concurrent_reply', False):
+                        # Return immediately if this is a concurrent handler
+                        self._logger.debug("%s FUTURE CONCURRENT", msg.name)
+                        return
```
bmerry: I haven't tried running it yet, but this neglects to schedule the call to async_reply.

katcp/kattypes.py
```diff
@@ -923,6 +923,31 @@ def make_reply(msgname, types, arguments, major):
...
+    ...     @concurrent_reply
+    ...     @tornado.gen.coroutine
+    ...     def request_myreq(self, req):
+    ...         result = yield self.slow_operation()
+    ...         return (req, result)
+    ...
+
+    """
+
+    handler.concurrent_reply = True
```
bmerry: I don't think this approach will work, because if another decorator (e.g. `@return_reply` in this case) is used to wrap the handler, the `concurrent_reply` attribute won't exist on the wrapper. In Python 3 this could be fixed by tracing the wrapping chain via `.__wrapped__`, but Python 2's `functools.wraps` doesn't set this. It could be worked around by having all the katcp wrappers propagate the attribute, but it will still break mysteriously if the user adds a wrapper.

brickZA: Hmm, I'm pretty sure `@wraps` updates the wrapped function object's `__dict__`  from the original one. So this would be a problem if the other decorates do not do this. Let me have a look, those decorators are quite old and may pre-date the fashion of using `wraps` :)

bmerry: It copies specific attributes (magic ones like `__doc__`) and you can give it a list of other ones to copy, but I don't think it copies everything.

By the way, you might be interested in the decorator Python module on PyPI. It creates signature-preserving decorators, which would simplify a lot of the trickery with `_orig_argnames`.

katcp/server.py
brickZA: Indeed, have now shuffled that line to the correct place :)

katcp/kattypes.py
brickZA: Refactoring those decorators to use the `decorator` module would simply that code considerably, but for now I'd rather to a minimal change without introducing a new dependency. Also I'm not that much of a fan of `decorator`'s exec trickery used to preserve the signature, but those concerns may be unfounded :)

For now I'm just making sure all the decorators use `wraps` as appropriate. See newest commit.

bmerry: Agreed, adding a new dependency is probably not something to conflate with this feature request.

I've made a simple test, and as I suspected, it doesn't work unless `@concurrent_reply` is the outermost decorator. The approach in #178 works with concurrent_reply inside the other decorators.

katcp/kattypes.py
```diff
@@ -698,8 +698,7 @@ def raw_handler(self, msg):
                 new_args = unpack_types(types, msg.arguments, argnames, major)
                 return handler(self, *new_args)
 
-        raw_handler.__name__ = handler.__name__
```
bmerry: `wraps` is designed as a `@` decorator; this could (arguably) be more readably written as `update_wrapper(raw_handler, handler)`. Similarly for other cases where `wraps` is not used as a decorator.

katcp/kattypes.py
```diff
@@ -709,7 +708,7 @@ def raw_handler(self, msg):
 
 # partial calls below 'copy' the function, letting us change the docstring without
 # affecting the original function's docstring
-inform = partial(request, has_req=False)
+inform = wraps(request)(partial(request, has_req=False))
 inform.__doc__ = """Decorator for inform handler methods.
```
bmerry: Might want to override `__name__` here and on unpack_message.

katcp/kattypes.py
```diff
@@ -923,6 +926,31 @@ def make_reply(msgname, types, arguments, major):
...
+    ...     @return_reply(Int())
+    ...     @concurrent_reply
+    ...     @tornado.gen.coroutine
+    ...     def request_myreq(self, req):
+    ...         result = yield self.slow_operation()
+    ...         return (req, result)
+    ...
+
+    """
+
```
bmerry: Should this be `_concurrent_reply` (leading underscore) for consistency with `_request_decorated`?

katcp/server.py
```diff
@@ -1055,6 +1056,9 @@ def handle_request(self, connection, msg):
                     def async_reply(f):
                         try:
                             connection.reply(f.result(), msg)
+                            if getattr(handler, 'concurrent_reply', False):
+                                self._logger.debug(
+                                    "%s FUTURE CONCURRENT replied", msg.name)
```
bmerry: Any particular reason to print this only for concurrent-reply futures, rather than for all future replies?

brickZA: Not really, would probably make sense to log it also.

katcp/kattypes.py
brickZA: Agreed

katcp/kattypes.py
brickZA: Aah, yes, good point.

katcp/kattypes.py
brickZA: You are probably right. Will update.

katcp/kattypes.py
brickZA: Is this still the case after sprinkling `wrap`s?

brickZA:
@bmerry I will incorporate your suggestions bit later today.

katcp/kattypes.py
bmerry: It works with the wraps.

bmerry:
It all looks good to me. Do you want to add unit tests and user docs to this pull or a separate one?

brickZA:
@bmerry In how much of a hurry are you to have it in master? I would prefer to do it all at once but if you would like to start using it we could merge this one and then do the other changes separately?

bmerry:
No hurry at all. I've got a workaround for now, and I was only thinking of this as a nice-to-have for some time in the future.

fjoubert:
After this and #179 has been merged, we should consider releasing a new version.

brickZA:
@ajoubertza @fjoubert @bmerry or anyone else that is interested, I have added a test for the concurrent reply behaviour, and some tests for the Async server case in general (mainly by subclassing the existing server tests and just setting up an async device server)

This PR should now be ready for merging.

katcp/kattypes.py
```diff
@@ -923,6 +930,31 @@ def make_reply(msgname, types, arguments, major):
...
+    Examples
+    --------
+    >>> class MyDevice(DeviceServer):
+    ...     @return_reply(Int())
+    ...     @concurrent_reply
+    ...     @tornado.gen.coroutine
+    ...     def request_myreq(self, req):
+    ...         result = yield self.slow_operation()
+    ...         return (req, result)
```
ajoubertza: This example is a coroutine, so it should probably use `raise` not `return`.

katcp/kattypes.py
```diff
@@ -923,6 +930,31 @@ def make_reply(msgname, types, arguments, major):
...
 
+def concurrent_reply(handler):
+    """Decorator for concurrent async request handlers
+
+    By default async request handlers that return a Future are serialised
+    per-connection, i.e. until the most recent handler resolves its future, the
+    next message will not be read from the client stream. A handler decorated
+    with this decorator allows the next message to be read before it has
+    resolved its future, allowing multiple requests from a single client to be
+    handled concurrently. This is similar to raising AsyncReply.
```
fjoubert: If I am not mistaken, then this would be a clean katcp way to replace what we hacked for `activate_subarray` in https://github.com/ska-sa/katobs/pull/211/files#diff-6ed26d613abbf9f5bd7b5d6de341109fR1462?

katcp/test/test_server.py
```diff
@@ -835,9 +861,66 @@ def test_add_remove_sensors(self):
...
+                    slow_f.done(),
+                    'slow async request should not be complete yet')
+                # Now cancel the slow command
+                reply, _ = yield self.client.future_request(
+                    katcp.Message.request('cancel-slow-command'))
+                self.assertTrue(reply.reply_ok(),
+                                '?cancel-slow-command should succeed')
+                slow_reply, _ = yield slow_f
+                self.assertTrue(slow_reply.reply_ok(),
+                                '?slow-command should succeed after being canceled')
```
ajoubertza: /s/canceled/cancelled

katcp/kattypes.py
brickZA: If, presumably, you were calling that in a request handler and you would like to be able to yield on that inside the request handler without blocking other client requests, yes, it could work :)

So, how about an approving review or a merge? :)

brickZA: Oh, sorry, did not see @ajoubertza's change requests below. Stupid inconsistent github page auto-refresh.

katcp/test/test_server.py
```diff
@@ -835,9 +861,66 @@ def test_add_remove_sensors(self):
...
+                t2 = time.time()
+                self.assertTrue(
+                    t2 - t1 < 0.1*slow_wait_time,
+                    'Slow request should be cancelled almost immediately')
+            except Exception, exc:
```
martinslabber: Would be nice to go: Thank me when we port to py3.
`except Exception as exc:`

katcp/server.py
```diff
@@ -1079,8 +1085,14 @@ def async_reply(f):
...
                     # reply.add_done_callback(async_reply)
-                    self._logger.debug("%s FUTURE OK" % (msg.name,))
-                    return done_future
+
+                    if concurrent:
+                        # Return immediately if this is a concurrent handler
+                        self._logger.debug("%s FUTURE CONCURRENT OK", msg.name)
+                        return
+                    else:
+                        self._logger.debug("%s FUTURE OK" % (msg.name,))
```
martinslabber: Since your have changed the other debug this one would also be nice as.
`self._logger.debug("%s FUTURE OK", msg.name)`
And thanks for using the args on debug logs.

katcp/kattypes.py
brickZA: Fixed in pending commit

katcp/test/test_server.py
brickZA: Fixed in pending commit.

katcp/test/test_server.py
brickZA: Yes, we need to get used to this! Fixed in pending commit.

katcp/server.py
brickZA: Fixed in pending commit

brickZA:
OK, everything pushed and ready for final merge I hope :)

ludwigschwardt:
Let's see if I can add the proposed "straggler group wait" before the next release :-)

PR #178, opened 2017-01-12
Add concurrent_reply decorator (CB-1747)

bmerry:
This adds a decorator @concurrent_reply that can be used on a request
handler that returns a future, to indicate that the request should be
fully concurrent (within the connection). It does this by wrapping the
return value in a ConcurrentReply wrapper object, which is detected and
unwrapped by DeviceServer.handle_request.

It still needs more documentation and unit tests. This is an initial
proof-of-concept.

### DIFF
--- a/katcp/core.py
+++ b/katcp/core.py
@@ -756,6 +756,23 @@ class AsyncReply(Exception):
     pass
 
 
+class ConcurrentReply(object):
+    """Returned by request handlers to indicate that the request should be
+    allowed to execute concurrently with other requests on the connection.
+    It wraps the actual reply, which is typically a future.
+
+    Examples
+    --------
+    >>> class MyDevice(DeviceServer):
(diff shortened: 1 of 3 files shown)


bmerry:
@brickZA this is my suggested alternative approach to #177, which should be less fragile to decorators not propagating the function object attribute.

brickZA:
@bmerry Thanks for the alternative.

 I think that decorators _should_ propagate function attributes; this is good Python practice in general and the pattern of using decorators to set function attributes is used elsewhere in the katcp module. To my eye PR #177 is more Pythonic and therefore preferable, although I will admit the possibility of bias :)

bmerry:
You're right, `functools.wraps` does propagate attributes, and #178 works for me now. And it's a little simpler too, so I'll close this.

PR #180, opened 2017-01-30
Remove spurious KeyboardInterrupt catch in IOLoopThreadWrapper

brickZA:
Catching it there serves no purpose. The only way that
`future.result(timeout)` can raise a KeyboardInterrupt is if the main
thread is waiting on a future running on the ioloop in another thread.
If SIGINT is received, the KeyboardInterrupt should indeed propagate.

### DIFF
--- a/katcp/resource_client.py
+++ b/katcp/resource_client.py
@@ -1473,13 +1473,9 @@ class IOLoopThreadWrapper(object):
             return future.result(timeout)
         except TimeoutError:
             raise
-        except KeyboardInterrupt:
-            print('=== KeyboardInterrupt ===')
-            sys.exc_clear()
-
         except Exception:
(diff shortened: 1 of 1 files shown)


katcp/resource_client.py
```diff
@@ -1473,13 +1473,9 @@ def call_in_ioloop(self, fn, args, kwargs, timeout=None):
             return future.result(timeout)
         except TimeoutError:
             raise
-        except KeyboardInterrupt:
```
martinslabber: I would have prefer, but thats a total different discussion.:

> except KeyboardInterrupt:
>     raise

brickZA: @martinslabber That seems like a philosophical point, since the effect is the same. Perhaps tell me over coffee :)

PR #182, opened 2017-04-12
Add timeout hint protocol flag CB-569

brickZA:
Idea is that we want a server to explicitly tell a client that it can use the ?request-timeout-hint request without the client having to guess.

### DIFF
--- a/katcp/core.py
+++ b/katcp/core.py
@@ -700,20 +700,64 @@ class DeviceMetaclass(type):
         for name in dir(mcs):
             if not callable(getattr(mcs, name)):
                 continue
+            handler = getattr(mcs, name)
             if name.startswith("request_"):
                 request_name = convert_method_name("request_", name)
-                mcs._request_handlers[request_name] = getattr(mcs, name)
-                assert(mcs._request_handlers[request_name].__doc__ is not None)
+                if mcs.check_protocol(handler):
+                    mcs._request_handlers[request_name] = handler
+                    assert(handler.__doc__ is not None)
             elif name.startswith("inform_"):
                 inform_name = convert_method_name("inform_", name)
-                mcs._inform_handlers[inform_name] = getattr(mcs, name)
-                assert(mcs._inform_handlers[inform_name].__doc__ is not None)
+                if mcs.check_protocol(handler):
+                    mcs._inform_handlers[inform_name] = handler
+                    assert(handler.__doc__ is not None)
                 # There is a bit of a name colission between the reply_*
                 # convention and the server reply_inform() method
             elif name.startswith("reply_") and name != 'reply_inform':
                 reply_name = convert_method_name("reply_", name)
-                mcs._reply_handlers[reply_name] = getattr(mcs, name)
-                assert(mcs._reply_handlers[reply_name].__doc__ is not None)
+                if mcs.check_protocol(handler):
+                    mcs._reply_handlers[reply_name] = handler
+                    assert(handler.__doc__ is not None)
+
+    def check_protocol(mcs, handler):
+        """Return False if `handler` should be filtered"""
+        # One cannot know protocol flags at definition time for device clients.
+        return True
+
+class DeviceServerMetaclass(DeviceMetaclass):
+    """Specialisation of DeviceMetaclass for Device Servers
+
+    Adds functionality for protocol-flag based filtering of handlers
+
+    """
+
+    def check_protocol(mcs, handler):
+        """
+        True if the current server's protocol flags satisfy handler requirements
+
+        """
+        try:
+            protocol_info = mcs.PROTOCOL_INFO
+        except Exception:
+            import pdb ; pdb.set_trace()
+        protocol_version = (protocol_info.major, protocol_info.minor)
+        protocol_flags = protocol_info.flags
+
+        # Check if minimum protocol version requirement is met
+        min_protocol_version = getattr(handler, '_minimum_katcp_version', None)
+        protocol_version_ok = (min_protocol_version is None or
+                               protocol_version >= min_protocol_version)
+
+        # Check if required optional protocol flags are present
+        required_katcp_protocol_flags = getattr(
+            handler, '_has_katcp_protocol_flags', None)
+        protocol_flags_ok = (
+            required_katcp_protocol_flags is None or
+            all(flag in protocol_flags
+                for flag in required_katcp_protocol_flags))
+
+        return protocol_version_ok and protocol_flags_ok
+
 
 
 class KatcpDeviceError(Exception):
--- a/katcp/core.py
+++ b/katcp/core.py
@@ -596,6 +596,9 @@ class ProtocolFlags(object):
 
     MULTI_CLIENT = 'M'
     MESSAGE_IDS = 'I'
+    # New proposal flag to indicate that a device supports ?request-timeout-hint
+    # See CB-2051
+    REQUEST_TIMEOUT_HINTS = 'T'
 
     STRATEGIES_V4 = frozenset(['none', 'auto', 'period', 'event',
                                'differential'])
@@ -607,15 +610,24 @@ class ProtocolFlags(object):
         5: STRATEGIES_V5
         }
 
+    REQUEST_TIMEOUT_HINTS_MIN_VERSION = (5, 1)
+
     def __init__(self, major, minor, flags):
         self.major = major
         self.minor = minor
         self.flags = set(list(flags))
         self.multi_client = self.MULTI_CLIENT in self.flags
         self.message_ids = self.MESSAGE_IDS in self.flags
+        self.request_timeout_hints = self.REQUEST_TIMEOUT_HINTS in self.flags
         if self.message_ids and self.major < MID_KATCP_MAJOR:
             raise ValueError(
                 'MESSAGE_IDS is only supported in katcp v5 and newer')
+        version_supports_hints = ((self.major, self.minor) >=
+                                  self.REQUEST_TIMEOUT_HINTS_MIN_VERSION)
+        if self.request_timeout_hints and not version_supports_hints:
+            raise ValueError(
+                'REQUEST_TIMEOUT_HINTS only suported in katcp v{}.{} and newer'
+                .format(self.major, self.minor))
 
     def strategy_allowed(self, strategy):
         return strategy in self.STRATEGIES_ALLOWED_BY_MAJOR_VERSION[self.major]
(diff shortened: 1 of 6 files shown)


katcp/core.py
```diff
@@ -596,6 +596,9 @@ class ProtocolFlags(object):
 
     MULTI_CLIENT = 'M'
     MESSAGE_IDS = 'I'
+    # New proposal flag to indicate that a device supports ?request-timeout-hint
+    # See CB-2051
+    REQUEST_TIMEOUT_HINTS = 'T'
```
brickZA: Done in pending commit

brickZA:
@ajoubertza At the moment the only actual implementation of ?request-timeout-hint is in proxy_base.py in katcore, where the request handler docstring describes it completely. To define it properly I suggest we try and add it to the KATCP spec, which would provide the definitive definition. Then we could probably also add an optional implementation (with proper docstring) to the top level DeviceServer class (only activated if the correct protocol flags are specified)

ajoubertza:
@brickZA Yes, it is nicely described in katcore, but that isn't a public repo yet.  If we add the flag here, we have to describe its purpose, so that it can be used.  A optional implementation would be great.

brickZA:
I agree, but I am willing to postpone that work :) Hence the TODO comment. I also made a backlog item in JIRA. Arm is available for twisting though.

ajoubertza:
I'd like to hear some other opinions.  @lvdheever?

brickZA:
Actually, let me just see how much effort it would be to port the katcore implementation to katcp

lvdheever:
Agreed that it would be great if it can be moved to katcp to wrapt this up. The problem with these "TODO"s and related JIRA tickets is that they never get enough priority to get attention. So I am against leaving little bits of TODOs for later. Rather just do it now if it is not a substantial amount of work.

brickZA:
I have come up with a nice clean way of doing it, but I will need some bits from https://github.com/ska-sa/katcp-python/pull/177

I only need to add some tests to PR177, so I'll go ahead and do that and then we can merge it and go forward.

katcp/core.py
```diff
@@ -607,15 +611,24 @@ class ProtocolFlags(object):
...
+        self.request_timeout_hints = self.REQUEST_TIMEOUT_HINTS in self.flags
         if self.message_ids and self.major < MID_KATCP_MAJOR:
             raise ValueError(
                 'MESSAGE_IDS is only supported in katcp v5 and newer')
+        version_supports_hints = ((self.major, self.minor) >=
+                                  self.REQUEST_TIMEOUT_HINTS_MIN_VERSION)
+        if self.request_timeout_hints and not version_supports_hints:
+            raise ValueError(
+                'REQUEST_TIMEOUT_HINTS only suported in katcp v{}.{} and newer'
+                .format(self.major, self.minor))
```
martinslabber: Please correct reported version info in error & s/suported/supported/.
```
raise ValueError(
      'REQUEST_TIMEOUT_HINTS only supported in katcp v{}.{} and newer'
       .format(*self.REQUEST_TIMEOUT_HINTS_MIN_VERSION)
```

katcp/core.py
```diff
@@ -687,20 +700,64 @@ def __init__(mcs, name, bases, dct):
         for name in dir(mcs):
             if not callable(getattr(mcs, name)):
                 continue
+            handler = getattr(mcs, name)
             if name.startswith("request_"):
                 request_name = convert_method_name("request_", name)
-                mcs._request_handlers[request_name] = getattr(mcs, name)
-                assert(mcs._request_handlers[request_name].__doc__ is not None)
+                if mcs.check_protocol(handler):
```
martinslabber: To be more DRY you could move the `if mcs.check_protocol(handler):` and assert out of the request/inform/reply condition.

katcp/core.py
```diff
@@ -687,20 +700,64 @@ def __init__(mcs, name, bases, dct):
...
+
+    def check_protocol(mcs, handler):
+        """
+        True if the current server's protocol flags satisfy handler requirements
+
+        """
+        try:
+            protocol_info = mcs.PROTOCOL_INFO
+        except Exception:
+            import pdb ; pdb.set_trace()
```
martinslabber: Seem you are still busy here.

brickZA: aah, thanks for catching that :)

katcp/core.py
brickZA: yes, but then you need an additional if statement to check if any of the three cases were triggered, etc. I agonised about it for _minutes_ ;)

brickZA:
@ajoubertza @martinslabber @lvdheever OK, now ready for re-review. Added default timeout hint request implementation and some machinery for including/excluding request handlers on the basis of the enabled KATCP server protocol flags / protocol version.

katcp/core.py
brickZA: Fixed and pushed.

katcp/test/test_server.py
```diff
@@ -275,8 +285,16 @@ class DeviceTestServerWrong(DeviceTestServer):
...
             DeviceTestServerWrong('', 0)
 
+katcp_version = __version__
 
 class test_DeviceServer(unittest.TestCase, TestUtilMixin):
+
+    expected_connect_messages = (
+            r'#version-connect katcp-protocol 5.0-IM',
+            # Will have to be updated for every library version bump
+            r'#version-connect katcp-library katcp-python-%s' % katcp_version,
```
ajoubertza: Does this comment still apply, if you're pulling the version from `__version__`, or are you referring to the `5.0-IM` line?

katcp/test/test_server.py
```diff
@@ -325,13 +338,91 @@ def test_has_sensor(self):
...
+        # Check that a request without a hint gives a timeout of 0 if explicitly
+        # asked for
+        req = mock_req('request-timeout-hint', 'help')
+        handle_mock_req(self.server, req)
+        self._assert_msgs_equal([req.reply_msg],
+                                ['!request-timeout-hint ok 1'])
+        self._assert_msgs_equal(req.inform_msgs,
+                                ['#request-timeout-hint help 0.0'])
+
+        # Now set hint on help command and check that it is reflected Note, in
```
ajoubertza: The "Note" shouldn't be capitalised, or is a full stop missing?

katcp/test/test_server.py
```diff
@@ -918,6 +1009,111 @@ def do_async_request_test(threadsafe_future, tornado_future):
...
+                         expected_handlers)
+        self.assertEqual(set(), actual_device_handlers & not_expected_handlers)
+
+
+    def test_handler_protocol_filters_all(self):
+        """Test handler filtering where everything should be included"""
+        self._test_handler_protocol_filters(self.DeviceWithEverything,
+                                            self.all_expected_handlers)
+
+    def test_handler_protocol_filters_five_with_nothing(self):
```
ajoubertza: Shouldn't this be a `five_one`, not a `five`?

katcp/test/test_server.py
```diff
@@ -918,6 +1009,111 @@ def do_async_request_test(threadsafe_future, tornado_future):
...
+
+        self._test_handler_protocol_filters(
+            DeviceVersionFiveOneWithNothing,
+            ['simple', 'version-5', 'version-51'])
+
+    def test_handler_protocol_filters_five_one_with_multi(self):
+        """Test handler filtering where protocol flags and version exclude some"""
+
+
+        class DeviceVersionFiveOneWithMultiClient(self.DeviceWithEverything):
```
ajoubertza: Too many blank lines, before this class?

katcp/test/test_server.py
```diff
@@ -918,6 +1009,111 @@ def do_async_request_test(threadsafe_future, tornado_future):
...
+        expected_handlers = set(expected_handlers)
+        not_expected_handlers = self.all_expected_handlers - expected_handlers
+        actual_device_handlers = set(DeviceCls._request_handlers.keys())
+
+        self.assertEqual(actual_device_handlers & expected_handlers,
+                         expected_handlers)
+        self.assertEqual(set(), actual_device_handlers & not_expected_handlers)
+
+
+    def test_handler_protocol_filters_all(self):
```
ajoubertza: Too many blank line before this method?

katcp/kattypes.py
```diff
@@ -956,6 +957,104 @@ def concurrent_reply(handler):
...
+
+    """
+    def decorator(handler):
+        handler._has_katcp_protocol_flags = protocol_flags
+        return handler
+
+    return decorator
+
+def request_timeout_hint(timeout_hint):
+    """Add recommended client timeout hint to a request for request
```
brickZA: Added in pending commit.

katcp/kattypes.py
```diff
@@ -956,6 +957,104 @@ def concurrent_reply(handler):
...
+
+    Useful for requests that take longer than average to reply. Hint is provided
+    to clients via ?request-timeout-hint. Note this is only exposed if the
+    device server sets the protocol version to KATCP v5.1 or higher and enables
+    the REQUEST_TIMEOUT_HINTS flag in its PROTOCOL_INFO class attribute
+
+    Parameters
+    ----------
+    timeout_hint : float (seconds) or None
+        How long the decorated request should reasonably take to reply
```
brickZA: Added in pending commit

katcp/kattypes.py
```diff
@@ -956,6 +957,104 @@ def concurrent_reply(handler):
...
+
+    Examples
+    --------
+    >>> class MyDevice(DeviceServer):
+    ...     @return_reply(Int())
+    ...     @request_timeout_hint
```
brickZA: Fixed in pending commit.

katcp/test/test_server.py
brickZA: I think you are right, it does not apply. The 5.0 bit should stay the same since that is the specified protocol version, although the library default might change at some stage.

brickZA: Clarified in pending commit.

katcp/test/test_server.py
brickZA: Punctuation updated in pending commit. I appreciate the attention to detail / level of pedantry, whichever way you prefer to look at it :)

katcp/test/test_server.py
brickZA: Fixed in pending commit.

katcp/test/test_server.py
brickZA: Corrected in pending commit.

katcp/test/test_server.py
brickZA: Made whitespace consistent between these related test methods.

brickZA:
@ajoubertza OK, updates pushed and ready for renewed scrutiny :)

PR #183, opened 2017-04-18
Move IOLoopThreadWrapper to ioloop_manager.py, a more sensible location

brickZA:


katcp/ioloop_manager.py
```diff
@@ -158,3 +165,84 @@ def join(self, timeout=None):
...
+            tornado_future.result()
+            # Should never get here since the tornado future should raise
+            assert False, 'Tornado Future should have raised'
+
+    def decorate_callable(self, callable_):
+        """Decorate a callable to use call_in_ioloop"""
+        @wraps(callable_)
+        def decorated(*args, **kwargs):
+            # Extract timeout from request itself or use default for ioloop wrapper
+            timeout = kwargs.get('timeout')
```
martinslabber: Do we need to remove timeout from the kwargs?
`timeout = kwargs.pop('timeout', None)`

katcp/test/test_ioloop_manager.py
```diff
@@ -0,0 +1,59 @@
...
+class test_ThreadsafeMethodAttrWrapper(unittest.TestCase):
+    def setUp(self):
+        self.ioloop_manager = ioloop_manager.IOLoopManager(managed_default=True)
+        self.ioloop = self.ioloop_manager.get_ioloop()
+        self.ioloop_thread_wrapper = ioloop_manager.IOLoopThreadWrapper(self.ioloop)
+        start_thread_with_cleanup(self, self.ioloop_manager, start_timeout=1)
+
+    def test_wrapping(self):
+        test_inst = self
+        class Wrappee(object):
```
martinslabber: :)

katcp/ioloop_manager.py
ludwigschwardt: What will happen if the ```callable_``` expects a timeout and then don't get it via ```kwargs``` anymore? It will probably revert to its own default, which seems worse than getting the intended timeout (even though it seems wrong to handle it twice).

brickZA: I just moved the class, I wasn't planning on opening a can of worms ;)

I think the rationale is that the most likely reason for adding a timeout argument would be to support the ioloop-bouncy-timeout. If you then leave in the kwarg for an original callable that does not take a `timeout` kwarg you would get an exception. Perhaps it would make sense to rename the kwarg to something less likely to conflict with the original?

fjoubert:
Bad news, this PR breaks everything 😭  see http://ci.camlab.kat.ac.za/job/IntegrationTests-KAT7-Deploy/661/console.

We are importing the `TimeoutError` incorrectly now from other packages. Let me know if you need some help with this, I suggest doing a complete kat-update and run some integration tests to see what else has been broken.

brickZA:
@fjoubert Try https://github.com/ska-sa/katcorelib/pull/236 for size to fix TimeoutError breakage.

katcp/ioloop_manager.py
ludwigschwardt: I would say that it is unlikely that the callable does not take a ```timeout``` kwarg - this way the timeout is an optional thing for the decorator. I don't think someone should send in ```timeout``` for the wrapper's sake...

PR #186, opened 2017-04-20
Let ClientGroup wait for a quorum of clients

ludwigschwardt:
Give ```ClientGroup.wait()``` the ability to wait until a subset of clients (a 'quorum') satisfy a sensor condition. This provides a way to e.g. ignore a broken or stowed receptor while slewing the working receptors onto a target. The wait times out if a quorum could not be achieved in the given timeout period. The desired quorum can be specified as either a number or a fraction of clients.

An additional grace period, which forms part of the overall timeout, allows for some fine-tuning of the behaviour. All remaining clients that have not yet satisfied the condition (the 'stragglers') are given the opportunity to do so within the grace period. This can effectively result in a full quorum by the time the wait resolves, if all clients behave as expected.

The default behaviour is unchanged: the group waits for all clients with a timeout of 5 seconds.

This also introduces the ```until_some``` helper function which returns a future that resolves when *some* of the passed futures resolve, as opposed to *all* (the usual yield on multiple futures) or *any* (the ```until_any``` function).

This addresses JIRA ticket [MKAIV-534](https://skaafrica.atlassian.net/browse/MKAIV-534).

### DIFF
--- a/katcp/core.py
+++ b/katcp/core.py
@@ -1805,7 +1805,7 @@ def until_later(delay, ioloop=None):
     return f
 
 def until_any(*futures, **kwargs):
-    """Return a future that resoves when any of the passed futures resolves.
+    """Return a future that resolves when any of the passed futures resolves.
 
     Resolves with the value yielded by the first future to resolve.
 
@@ -1898,3 +1898,48 @@ def future_timeout_manager(timeout=None, ioloop=None):
     maybe_timeout.remaining = _remaining
 
     return maybe_timeout
+
+@tornado.gen.coroutine
+def until_some(*args, **kwargs):
+    """Return a future that resolves when some of the passed futures resolve.
+
+    The futures can be passed as either a sequence of *args* or a dict of
+    *kwargs* (but not both). Some additional keyword arguments are supported,
+    as described below. Once a specified number of underlying futures have
+    resolved, the returned future resolves as well, or a timeout could be
+    raised if specified.
+
+    Parameters
+    ----------
+    done_at_least : None or int
+        Number of futures that need to resolve before this resolves (default all)
+    timeout : None or float
+        Timeout in seconds, or None for no timeout (the default)
+
+    Returns
+    -------
+    This command returns a tornado Future that resolves with a list of
+    (index, value) tuples containing the results of all futures that resolved,
+    with corresponding indices (numbers for *args* futures or keys for *kwargs*
+    futures).
+
+    Raises
+    ------
+    :class:`tornado.gen.TimeoutError`
+        If operation times out before the requisite number of futures resolve
+
+    """
+    done_at_least = kwargs.pop('done_at_least', None)
+    timeout = kwargs.pop('timeout', None)
+    # At this point args and kwargs are either empty or contain futures only
+    if done_at_least is None:
+        done_at_least = len(args) + len(kwargs)
+    wait_iterator = tornado.gen.WaitIterator(*args, **kwargs)
+    maybe_timeout = future_timeout_manager(timeout)
+    results = []
+    while not wait_iterator.done():
+        result = yield maybe_timeout(wait_iterator.next())
+        results.append((wait_iterator.current_index, result))
+        if len(results) >= done_at_least:
+            break
+    raise tornado.gen.Return(results)
--- a/katcp/core.py
+++ b/katcp/core.py
@@ -1912,7 +1912,8 @@ def until_some(*args, **kwargs):
     Parameters
     ----------
     done_at_least : None or int
-        Number of futures that need to resolve before this resolves (default all)
+        Number of futures that need to resolve before this resolves or None
+        to wait for all (default None)
     timeout : None or float
         Timeout in seconds, or None for no timeout (the default)
 
--- a/katcp/resource_client.py
+++ b/katcp/resource_client.py
@@ -5,6 +5,7 @@ import logging
 import sys
 import re
 import collections
+import math
 import time
 
 import tornado
@@ -20,7 +21,7 @@ from katcp.resource import KATCPReply, KATCPSensorError
 from katcp.core import (AttrDict, DefaultAttrDict, AsyncCallbackEvent,
                         steal_docstring_from,
                         AsyncState, AsyncEvent, LatencyTimer,
-                        until_any, log_future_exceptions)
+                        until_any, until_some, log_future_exceptions)
 
 # TODO NM 2017-04-13 Importing IOLoopThreadwrapper here for backwards
 # compatibility, user code should be changed to import it from the more logical
@@ -1085,7 +1086,8 @@ class ClientGroup(object):
         raise tornado.gen.Return(sensors_strategies)
 
     @tornado.gen.coroutine
-    def wait(self, sensor_name, condition_or_value, timeout=5):
+    def wait(self, sensor_name, condition_or_value, timeout=5.0, quorum=None,
+             max_grace_period=1.0):
         """Wait for sensor present on all group clients to satisfy a condition.
 
         Parameters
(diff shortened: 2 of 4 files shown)


katcp/resource_client.py
```diff
@@ -1114,16 +1124,37 @@ def wait(self, sensor_name, condition_or_value, timeout=5):
             sensor is not present
 
         """
+        if quorum is None:
+            quorum = len(self.clients)
+        elif quorum < 1:
+            quorum = int(math.ceil(quorum * len(self.clients)))
```
ajoubertza: What happens if `quorum` is exactly 1?  We're not sure if the user intended to wait for 100% of the clients, or just for 1 of them.  As is, the code will assume they meant to wait for 1 client.  Maybe this just needs to be clearer in the docstring.  If they want to wait for 100%, they must use `None`.

katcp/resource_client.py
```diff
@@ -1114,16 +1124,37 @@ def wait(self, sensor_name, condition_or_value, timeout=5):
...
-        results = yield futures
+            f = client.wait(sensor_name, condition_or_value, initial_timeout)
+            futures[client.name] = f
+        # No timeout required here as all futures will resolve after timeout
+        initial_results = yield until_some(done_at_least=quorum, **futures)
+        results = dict(initial_results)
+        # Identify stragglers and let them all respond within grace period
+        stragglers = {}
+        for client in self.clients:
+            if not results.get(client.name, False):
```
ajoubertza: Splitting hairs, but
```
if client.name not in results:
```
is a little more readable for me.

katcp/resource_client.py
```diff
@@ -1114,16 +1124,37 @@ def wait(self, sensor_name, condition_or_value, timeout=5):
...
+                f = client.wait(sensor_name, condition_or_value, grace_period)
+                stragglers[client.name] = f
+        rest_of_results = yield until_some(**stragglers)
+        results.update(dict(rest_of_results))
```
ajoubertza: Some Tornado newb questions:

Does Tornado just silently drop the futures that haven't resolved yet when their original timeouts expire (we may have exited this method by then)?

We are calling `client.wait` a second time on the same sensor for the stragglers - is KATCP happy to have multiple waits on the same sensor?  I guess the async design would be fine with this.

Assuming a straggler succeeds in the grace period, then both futures will have resolved at that point, right?

katcp/core.py
```diff
@@ -1898,3 +1898,48 @@ def maybe_timeout(f):
...
+    if done_at_least is None:
+        done_at_least = len(args) + len(kwargs)
+    wait_iterator = tornado.gen.WaitIterator(*args, **kwargs)
+    maybe_timeout = future_timeout_manager(timeout)
+    results = []
+    while not wait_iterator.done():
+        result = yield maybe_timeout(wait_iterator.next())
+        results.append((wait_iterator.current_index, result))
+        if len(results) >= done_at_least:
+            break
```
martinslabber: WaitIterator iterates through the futures as you pass them in [1].
Thus if you have a timeout on the first future (even with maybe_timeout) and all the other were successful it would still result in a future exception. If thats the case you have to catch timeouts and reraise outside the while loop.

[1] Hope to be wrong then it is much simpler for you.

I'll try explaining in code - something like.
```
    your_clear_to_loanch_the_timeout_moab = 0
    while not wait_iterator.done():
         try:
             result = yield maybe_timeout(wait_iterator.next())
         except tornado.gen.TimeoutError:
             your_clear_to_loanch_the_timeout_moab += 1
         results.append((wait_iterator.current_index, result))
         if len(results) >= done_at_least:
             break
    if your_clear_to_loanch_the_timeout_moab:
         raise tornado.gen.TimeoutError("timeout")
```
--------------------
From the code I assume your primary use case is to wait for a handful of wait-until-sensor-value-reaches(X) futures. I like the general nature of this but think for requests it will need to be done slightly differently, taking into account the result.
P.S. I am away for a few days, so @brickZA can decide.

katcp/resource_client.py
ludwigschwardt: Indeed. Or you can pick 0.999 as it rounds up :-)

I've improved the docstring wording in a pending commit.

katcp/resource_client.py
ludwigschwardt: That's actually not the same thing... I had it like that and then realised I want to revisit all clients with either False results or with no results. The former occurs in the midst of an initial timeout, where at least one future will resolve with False. These need to be revisited too.

katcp/resource_client.py
ludwigschwardt: Some Tornado newb answers:

All of these client.wait futures will resolve as that's how they are constructed. If the sensor condition is True, resolve with True, otherwise wait and then upon timeout resolve with False. See katcp/resource.py around line 217. So they will sit on the ioloop until such time if I have to guess.

I'm guessing multiple waits on the same sensor will be fine too.

And yes, both futures will resolve in your last statement.

katcp/core.py
ludwigschwardt: [1] You are wrong so it is much simpler :-)

> (WaitIterator:) Provides an iterator to yield the results of futures *as they finish*.
> [...]
> If you need to get the result of each future *as soon as possible*, or if you need the result of some futures *even if others produce errors*, you can use WaitIterator.
> [...]
> Because results are returned as soon as they are available the output from the iterator *will not be in the same order as the input arguments*.

If WaitIterator had worked as you described it would be rather useless... You might as well have done (except for parallelism):

```Python
results = []
for future in futures:
    result = yield future
    results.append(result)
```

and all the exception handling becomes your baby.

ludwigschwardt: I'd like to hear more about the requests use case...

ludwigschwardt:
Coming back to the data issue: It simply means that some antennas should be flagged for some periods of time in the data. We know this through the fabled "CAM flag" that does not quite exist yet (see issue [SR-730](https://skaafrica.atlassian.net/browse/SR-730) and this [doc](https://docs.google.com/a/ska.ac.za/document/d/1KcGRCHGX9CVu5u5KfnlH9Rl3S7_S4b4MimU2HBBqK2Q/edit?usp=sharing)).

katcp/resource_client.py
ajoubertza: Oh, I see now.  Thanks.

katcp/resource_client.py
ajoubertza: Thanks for the details.

brickZA: Tornado does not support cancelling futures, so potentially you could be building up a bunch of futures that never resolve. In the case of using a coroutine (and some other mechanisms that I cannot remember off the top of my head), the ioloop itself will add a done callback to the future. This means that the ioloop (which is forever) has a reference to the future via the done callback. When a future is resolved it deletes all the done callbacks, so als deletes the ioloop -> future reference, allowing the future to be collected. If the future does _not_ resolve, it will not be collected.

We have in the past had an issue where we were waiting for a set of possible states in a loop using `until_any`() where each state was represented using a separate future. Each time the state waiting was done, a new set of futures were creates. When any of the states occured, the `until_any()` future would resolve, but not the states that were not reached. Those states were waited upon by the ioloop, so they would only be collected if they resolved. Some of the states happened very rarely, so copies of those futures built up with each iteration of the loop.

A bodgey hackey way of dealing with that was to have 'until_any()' reach into the future's private members and delete the  done  callbacks on the unresolved futures.

We _might_ run into a similar issue with `until_some()`, but since we are likely to use it only for things that are likely to happen within a fixed time and also in 'short-lived' observation scripts, it seemed better not to hack it. Hopefully tornado futures will one day allow themselves to be cancelled or we will switch over to another async framework :)

katcp/core.py
```diff
@@ -1898,3 +1898,48 @@ def maybe_timeout(f):
...
+    The futures can be passed as either a sequence of *args* or a dict of
+    *kwargs* (but not both). Some additional keyword arguments are supported,
+    as described below. Once a specified number of underlying futures have
+    resolved, the returned future resolves as well, or a timeout could be
+    raised if specified.
+
+    Parameters
+    ----------
+    done_at_least : None or int
+        Number of futures that need to resolve before this resolves (default all)
```
ludwigschwardt: Done in pending commit.

katcp/test/test_core.py
```diff
@@ -490,3 +490,42 @@ def call_set():
...
+        f3 = tornado.concurrent.Future()
+        # Test for timing out
+        with self.assertRaises(tornado.gen.TimeoutError):
+            yield until_some(f1, f2, f3, timeout=0.05)
+        # Test for NOT timing out
+        f3.set_result(84)
+        f1.set_result(24)
+        f2.set_result(42)
+        results = yield until_some(f1, f2, f3, timeout=0.1)
+        self.assertEqual(results, [(0, 24), (1, 42), (2, 84)],
```
ludwigschwardt: Smoothed the rough edges in a pending commit. Put generous warnings that 1.0 != 1 .

The futures are coming out in index order if they were resolved before WaitIterator starts, but it is an implementation feature, so I'll modify accordingly.

PR #187, opened 2017-05-05
User/nmarais/ci 27/initial futurize work for python 2/3 compatibility

brickZA:
Futurize a small part of KATCP lib as a proof of concept. Also got rid of the python 2 ProxyTypes dependency by moving it into our codebase (only about 200 lines) and also making it python 2/3 compatible.

Going forward, files that have been marked with `# Python 2/3 compatibility stuff` comment near the top should use python 2/3 compatible idioms: http://python-future.org/compatible_idioms.html

### DIFF
--- a/katcp/object_proxies.py
+++ b/katcp/object_proxies.py
@@ -1,0 +1,183 @@
+# Copied from peak.util.proxies https://pypi.python.org/pypi/ProxyTypes/0.9
+# to allow translation to python 2/3 compatible code # According to the pypi
+# index record this code is distributed under PSF (Python Software License) or
+# ZPL (Zope Public License if I' not mistaken). The PSF license is BSD style and
+# allows us to copy, modify and redistribute this code.
+
+class AbstractProxy(object):
+    """Delegates all operations (except ``.__subject__``) to another object"""
+    __slots__ = ()
+
+    def __call__(self,*args,**kw):
+        return self.__subject__(*args,**kw)
+
+    def __getattribute__(self, attr, oga=object.__getattribute__):
+        subject = oga(self,'__subject__')
+        if attr=='__subject__':
+            return subject
+        return getattr(subject,attr)
+
+    def __setattr__(self,attr,val, osa=object.__setattr__):
+        if attr=='__subject__':
+            osa(self,attr,val)
+        else:
+            setattr(self.__subject__,attr,val)
+
+    def __delattr__(self,attr, oda=object.__delattr__):
+        if attr=='__subject__':
+            oda(self,attr)
+        else:
+            delattr(self.__subject__,attr)
+
+    def __nonzero__(self):
+        return bool(self.__subject__)
+
+    def __getitem__(self,arg):
+        return self.__subject__[arg]
+
+    def __setitem__(self,arg,val):
+        self.__subject__[arg] = val
+
+    def __delitem__(self,arg):
+        del self.__subject__[arg]
+
+    def __getslice__(self,i,j):
+        return self.__subject__[i:j]
+
+
+    def __setslice__(self,i,j,val):
+        self.__subject__[i:j] = val
+
+    def __delslice__(self,i,j):
+        del self.__subject__[i:j]
+
+    def __contains__(self,ob):
+        return ob in self.__subject__
+
+    for name in 'repr str hash len abs complex int long float iter oct hex'.split():
+        exec "def __%s__(self): return %s(self.__subject__)" % (name,name)
+
+    for name in 'cmp', 'coerce', 'divmod':
+        exec "def __%s__(self,ob): return %s(self.__subject__,ob)" % (name,name)
+
+    for name,op in [
+        ('lt','<'), ('gt','>'), ('le','<='), ('ge','>='),
+        ('eq','=='), ('ne','!=')
+    ]:
+        exec "def __%s__(self,ob): return self.__subject__ %s ob" % (name,op)
+
+    for name,op in [('neg','-'), ('pos','+'), ('invert','~')]:
+        exec "def __%s__(self): return %s self.__subject__" % (name,op)
+
+    for name, op in [
+        ('or','|'),  ('and','&'), ('xor','^'), ('lshift','<<'), ('rshift','>>'),
+        ('add','+'), ('sub','-'), ('mul','*'), ('div','/'), ('mod','%'),
+        ('truediv','/'), ('floordiv','//')
+    ]:
+        exec (
+            "def __%(name)s__(self,ob):\n"
+            "    return self.__subject__ %(op)s ob\n"
+            "\n"
+            "def __r%(name)s__(self,ob):\n"
+            "    return ob %(op)s self.__subject__\n"
+            "\n"
+            "def __i%(name)s__(self,ob):\n"
+            "    self.__subject__ %(op)s=ob\n"
+            "    return self\n"
+        )  % locals()
+
+    del name, op
+
+    # Oddball signatures
+
+    def __rdivmod__(self,ob):
+        return divmod(ob, self.__subject__)
+
+    def __pow__(self,*args):
+        return pow(self.__subject__,*args)
+
+    def __ipow__(self,ob):
+        self.__subject__ **= ob
+        return self
+
+    def __rpow__(self,ob):
+        return pow(ob, self.__subject__)
+
+
+class ObjectProxy(AbstractProxy):
+    """Proxy for a specific object"""
+
+    __slots__ = "__subject__"
+
+    def __init__(self,subject):
+        self.__subject__ = subject
+
+
+class CallbackProxy(AbstractProxy):
+    """Proxy for a dynamically-chosen object"""
+
+    __slots__ = '__callback__'
+
+    def __init__(self, func):
+        set_callback(self,func)
+
+set_callback = CallbackProxy.__callback__.__set__
+get_callback = CallbackProxy.__callback__.__get__
+CallbackProxy.__subject__ = property(lambda self, gc=get_callback: gc(self)())
+
+
+
+class LazyProxy(CallbackProxy):
+    """Proxy for a lazily-obtained object, that is cached on first use"""
+    __slots__ = "__cache__"
+
+get_cache = LazyProxy.__cache__.__get__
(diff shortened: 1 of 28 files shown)


ludwigschwardt:
I'm wondering what you folk make of this:

http://www.python3statement.org/

All major scientific Python packages are busy becoming Python 3 only. I'm wondering whether we should consider staying only briefly at the 2/3 compatibility stage as a crutch to get to 3-only...

bmerry:
Certainly I think as soon as we're able to run everything in Python 3 we should stop supporting Python 2 ASAP. The only packages where I think it makes sense to continue supporting Python 2 is our open-source packages, and in that case only until 2020.

I would certainly like to convert my trollius-using code to use asyncio instead, which would mean dropping Python 2 support.

brickZA:
@ludwigschwardt I agree that there is no point to maintain py2 compatibility (for CAM at least) as soon as we have py3 working, but the only practical way to get there (IMHO) is to go through the Py6 stage :)

katcp/resource_client.py
```diff
@@ -669,7 +676,7 @@ def _add_sensors(self, sensor_keys):
         sensor_instances = yield sensor_instance_fut
         # Store KATCPSensor instances in self.sensor
         added_names = []
-        for s_name, s_obj in sensor_instances.items():
+        for s_name, s_obj in list(sensor_instances.items()):
```
bmerry: 2to3 tends to be very conservative and makes these conversions to give identical behaviour to the old code.

In a lot of cases what you probably really want is
```python
for s_name, s_obj in six.iteritems(sensor_instances)
```
to iterate over the original rather than making a copy, but it needs to be done carefully: if the loop mutates the dictionary then it's not equivalent.

katcp/test/test_core.py
```diff
@@ -7,8 +7,12 @@
 """Tests for the katcp utilities module.
    """
 
+# Python 2/3 compatibility stuff
+from builtins import str
+from builtins import object
 import logging
-import unittest2 as unittest
+import unittest
```
bmerry: Have you checked that you're not using any unittest2 features?

katcp/test/test_resource_client.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import print_function
```
bmerry: Personally I always import print_function, division and absolute_import, to be more sure of getting consistent behaviour between Python 2 and 3. Division is particularly nasty and probably can't be handled by automatic tools since the typing is dynamic.

katcp/test/test_resource_client.py
```diff
@@ -16,7 +24,7 @@
 import tornado
 import mock
 
-from thread import get_ident as get_thread_ident
+from _thread import get_ident as get_thread_ident
```
bmerry: This looks a bit odd. Surely `_thread` is an implementation-detail module?

katcp/test/test_resource_client.py
```diff
@@ -1167,8 +1191,10 @@ def __eq__(self, other):
             self.assertEqual(getattr(wrapped_dict, key).wrappee,
                              getattr(test_dict, key))
         # Test whole dict comparison
-        self.assertEqual(wrapped_dict,
-                         {k : TestWrapper(v) for k, v in test_dict.items()})
+        # TODO PY3 check if we can get rid of the list() call
```
bmerry: In this case it looks very safe to do so.

katcp/test/test_resource_client.py
```diff
@@ -1245,7 +1271,8 @@ def setUp(self):
             'resource1' : dict(controlled=True),
             'resource2' : dict(controlled=True)},
                                  name='wraptest')
-        self.resource_names = self.default_spec['clients'].keys()
+        # TODO PY3 check if we can get rid of the list() call
```
bmerry: Here you probably don't want to, since then resource_names wouldn't be a list and could only be iterated once.

katcp/test/test_resource_client.py
```diff
@@ -1158,7 +1181,8 @@ def __eq__(self, other):
 
         wrapped_dict = resource_client.AttrMappingProxy(test_dict, TestWrapper)
         # Test keys
-        self.assertEqual(wrapped_dict.keys(), test_dict.keys())
+        # TODO PY3 check if we can get rid of the list() call
```
bmerry: Probably not, since comparing iterators won't make much sense.

katcp/test/test_resource_client.py
```diff
@@ -1132,14 +1153,16 @@ def test_group_wait(self):
...
+        for client in list(DUT.children.values()):
             if client.name == selected_client:
                 self.assertTrue(result[client.name])
             else:
                 self.assertFalse(result[client.name])
         result = yield group.wait('wait_sensor', 1, timeout=0.1, quorum=2)
         self.assertFalse(result)
-        for client in DUT.children.values():
+        # TODO PY3 check if we can get rid of the list() call
```
bmerry: Looks safe to do so here.

katcp/test/test_resource_client.py
```diff
@@ -1132,14 +1153,16 @@ def test_group_wait(self):
         self.assertFalse(result)
         result = yield group.wait('wait_sensor', 1, timeout=0.1, quorum=0.1)
         self.assertTrue(result)
-        for client in DUT.children.values():
+        # TODO PY3 check if we can get rid of the list() call
```
bmerry: Looks safe to do so here.

katcp/test/test_resource_client.py
```diff
@@ -1114,16 +1132,19 @@ def test_group_wait(self):
...
         self.assertFalse(result)
-        for client in DUT.children.values():
+        # TODO PY3 check if we can get rid of the list() call
+        for client in list(DUT.children.values()):
             self.assertFalse(result[client.name])
         # Test quorum functionality
         with self.assertRaises(TypeError):
             yield group.wait('wait_sensor', 1, timeout=0.1, quorum=1.1)
-        selected_client = DUT.children.keys()[0]
+        # TODO PY3 check if we can get rid of the list() call
```
bmerry: Could possibly replace by `next(DUT.children.keys())`, although the exception will be different if the dict is empty.

katcp/test/test_resource_client.py
```diff
@@ -1114,16 +1132,19 @@ def test_group_wait(self):
         result = yield group.wait('wait_sensor', 0, timeout=None, quorum=1.0)
         self.assertTrue(result)
         # Check detailed results per client
-        for client in DUT.children.values():
+        # TODO PY3 check if we can get rid of the list() call
```
bmerry: Looks safe.

katcp/test/test_resource_client.py
```diff
@@ -1114,16 +1132,19 @@ def test_group_wait(self):
...
+        for client in list(DUT.children.values()):
             self.assertTrue(result[client.name])
         result = yield group.wait('wait_sensor', 1, timeout=0.1)
         self.assertFalse(result)
-        for client in DUT.children.values():
+        # TODO PY3 check if we can get rid of the list() call
```
bmerry: Looks safe.

katcp/object_proxies.py
```diff
@@ -0,0 +1,187 @@
...
+            attr=='__subject__'
+            or hasattr(type(self),attr) and not attr.startswith('__')
+        ):
+            oda(self,attr)
+        else:
+            delattr(self.__subject__,attr)
+
+class ObjectWrapper(ObjectProxy, AbstractWrapper):      __slots__ = ()
+class CallbackWrapper(CallbackProxy, AbstractWrapper):  __slots__ = ()
+class LazyWrapper(LazyProxy, AbstractWrapper):          __slots__ = ()
```
martinslabber: Would be nice if you PEP8 this file. Its an ideal time to play with yapf.

brickZA:
@bmerry Comments noted about the `list()` calls. I feel at this stage, particularly since we don't have Py3 continuous integration set up yet and the potentially subtle nature of bugs that can be introduced, it is better to  be conservative. As you noted care has to be taken if the sequences being iterated over might mutated during the loop, and just making a copy by default will mitigate that.

I don't think the overhead of the list conversion will be particularly significant in any of these cases. I'd rather we make sure our code is equivalent to what it was before conversion and worry about removing unnecessary `list()` calls further down the line when we abandon Py2 support. In cases where we explicitly used `iteritems()` in the current codebase the `six.iteritems()` option would of course be appropriate.

katcp/object_proxies.py
brickZA: I thought about that too, but since it is basically a copy of an upstream module it seemed to make sense to leave it as close to original as possible. If, however, someone went to the effort of making a PEP8 PR for that I would not argue :)

katcp/test/test_core.py
brickZA: unittest2 is  a backport of Python 2.7's unittest to older versions. AFAIK the only difference between python 2.7+ unittest and unittest2 are some of the message printing defaults, but nothing substantial.

katcp/test/test_resource_client.py
brickZA: The translator tool does use an `old_div()` workaround if it doesn't know, but in the case where it did it I knew that at least one of the numbers was guaranteed to be a float, so I replaced with with a normal `/`.  But probably not a bad idea to do the imports as you suggest.

brickZA: OK, I did add it to all files, will probably help slightly to prevent introducing subtle Py2-isms until we are fully ported.

katcp/test/test_resource_client.py
brickZA: I dont' know, that is what the Futurize tool did, and it seems to work. So I assume that is the correct way?

bmerry: It looks like the `_thread` module is provided by the `future` package, so yes, this is presumably the right thing to do. In Python 3 it also seems to be available as `threading.get_ident` but that doesn't work on Python 2.

mmphego: @tockards see this comment regarding `_thread` import.

tockards: @mmphego thank you.

PR #188, opened 2017-06-21
Add a tornado AsyncWaitingMock complement to WaitingMock CB-1466

brickZA:
In the testutils module.

In support of implementing the async dev handler tests.

### DIFF
--- a/katcp/testutils.py
+++ b/katcp/testutils.py
@@ -18,6 +18,7 @@ import functools
 import mock
 import tornado.testing
 import tornado.ioloop
+import tornado.locks
 
 from thread import get_ident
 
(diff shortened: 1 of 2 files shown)


katcp/testutils.py
```diff
@@ -1809,6 +1810,39 @@ def assert_wait_call_count(self, count, timeout=1.):
...
+        t0 = time.time()
+        if self.call_count >= count and len(self.call_args_list) >= count:
+            raise tornado.gen.Return(True)
+        to_wait = timeout
+
+        while to_wait >= 0 and self.call_count < count:
+            try:
+                yield self._call_event.wait(timeout)
+                self._call_event.clear()
+            except gen.TimeoutError:
```
ajoubertza: Slightly inconsistent use of the tornado namespace: lines 1824 and 1833 use `tornado.gen`, while here we just use `gen` directly.  Interestingly, we don't have an `import tornado` or `import tornado.gen` line in this file, so is another import bringing `tornado.gen` into the namespace, or do I just misunderstand these things?

On a probably unrelated note, I see the Jenkins console output reports a few of these immediately after running the tests.
```
Exception AttributeError: "'NoneType' object has no attribute 'gen'" in <generator object connect at 0x7f75a490fd20> ignored
```
Any idea what that is?

katcp/testutils.py
```diff
@@ -1809,6 +1810,39 @@ def assert_wait_call_count(self, count, timeout=1.):
...
+
+        """
+        t0 = time.time()
+        if self.call_count >= count and len(self.call_args_list) >= count:
+            raise tornado.gen.Return(True)
+        to_wait = timeout
+
+        while to_wait >= 0 and self.call_count < count:
+            try:
+                yield self._call_event.wait(timeout)
```
martinslabber: yield self._call_event.wait(timeout) >> yield self._call_event.wait(to_wait)

For my edification: On line 1832 why are you also checking the len of self.call_args_list?

brickZA: Thanks, fixed

The second check is indeed not necessary, cut-n-paste error from the threaded-safe version where all kinds of fun race conditions occur around call_count and the call list :)

katcp/testutils.py
brickZA: - Will look at making the namespace use consistent (commit pending)
- Doing `from tornado import gen` will indeed bring tornado.gen into existence. It is not always apparent that by default importing a top level module (e.g. `import tornado`) does not automatically import submodules. However, many packages will import some of the submodules in the top-level module's `__init__.py`, clouding the issue.
- Will look into the error, it might be a missing `gen_test` decorator

brickZA: I'm not sure what the attribute errors are about. I don't see them when I run locally. It only happens after the actual nosetests are done, which makes me think that there may be some daemon (shudder) threads running at the end of the tests getting tangled up when the python process tears itself down. See https://joeshaw.org/python-daemon-threads-considered-harmful/

brickZA:
OK, ready for re-review and comittance :)

katcp/testutils.py
ajoubertza: OK.  I see now that those exceptions were there before your changes anyway.

PR #193, opened 2017-10-30
Added error handling when parsing Message arguments and remove unicode characters from devices that do not only send ASCII (not following spec).

fjoubert:
Also some minor pep8 fixes.

Added a simple test for a unicode message.

### DIFF
--- a/katcp/core.py
+++ b/katcp/core.py
@@ -299,7 +299,23 @@ class Message(object):
         elif isinstance(arg, bool):
             return str(int(arg))
         else:
-            return str(arg)
+            try:
+                return str(arg)
+            except Exception:
+                # unicode characters will break the str cast, so
+                # try to encode to ascii and replace the offending characters
+                # with a '?' character
+                logger.exception("Error casting message argument to str!")
+                try:
+                    if type(arg) != unicode:
+                        arg = arg.decode('utf-8')
+                    return arg.encode('ascii', 'replace')
+                except Exception:
+                    # If all else fails, ignore the string completely
+                    logger.exception(
+                        "Error decoding message argument from utf-8, "
+                        "ignoring message and returning ''")
+                    return ''
 
     def copy(self):
         """Return a shallow copy of the message object and its arguments.
@@ -611,7 +627,7 @@ class ProtocolFlags(object):
     STRATEGIES_ALLOWED_BY_MAJOR_VERSION = {
         4: STRATEGIES_V4,
         5: STRATEGIES_V5
-        }
+    }
 
     REQUEST_TIMEOUT_HINTS_MIN_VERSION = (5, 1)
 
@@ -638,8 +654,8 @@ class ProtocolFlags(object):
     def __eq__(self, other):
         if not isinstance(other, ProtocolFlags):
             return NotImplemented
-        return (self.major == other.major and self.minor == other.minor
-                and self.flags == other.flags)
+        return (self.major == other.major and self.minor == other.minor and
+                self.flags == other.flags)
 
     def __str__(self):
         flag_str = self.flags and ("-" + "".join(sorted(self.flags))) or ""
@@ -756,7 +772,6 @@ class DeviceServerMetaclass(DeviceMetaclass):
                 for flag in required_katcp_protocol_flags))
 
         return protocol_version_ok and protocol_flags_ok
-
 
 
 class KatcpDeviceError(Exception):
@@ -1006,7 +1021,7 @@ class Sensor(object):
             description = '%(type)s sensor %(name)r %(unit_description)s' % (
                           dict(type=self.stype.capitalize(), name=self.name,
                                unit_description=('in unit '+units if units else
-                                                 'with no unit')) )
+                                                 'with no unit')))
 
         self.description = description
         self.units = units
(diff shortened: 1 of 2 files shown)


katcp/core.py
```diff
@@ -299,7 +299,23 @@ def format_argument(self, arg):
         elif isinstance(arg, bool):
             return str(int(arg))
         else:
-            return str(arg)
+            try:
+                return str(arg)
+            except Exception:
```
bmerry: Should this rather catch only encoding errors? If it's some class that just doesn't support conversion to str at all, it seems like this is going to hide a real error.

katcp/core.py
```diff
@@ -299,7 +299,23 @@ def format_argument(self, arg):
...
+                # unicode characters will break the str cast, so
+                # try to encode to ascii and replace the offending characters
+                # with a '?' character
+                logger.exception("Error casting message argument to str!")
+                try:
+                    if type(arg) != unicode:
```
bmerry: `isinstance` is generally preferable to exact type comparisons.

katcp/core.py
```diff
@@ -299,7 +299,23 @@ def format_argument(self, arg):
...
+                        arg = arg.decode('utf-8')
+                    return arg.encode('ascii', 'replace')
+                except Exception:
+                    # If all else fails, ignore the string completely
+                    logger.exception(
+                        "Error decoding message argument from utf-8, "
```
bmerry: Not necessarily true. The failure could be in `arg.encode`, if `arg` isn't unicode.

katcp/core.py
fjoubert: Yeah thats a good point. Changing to catching `UnicodeEncodeError`

katcp/core.py
fjoubert: Fair, I knew this at some stage. Changing to use `isinstance`.

katcp/core.py
fjoubert: You're right, I changed it to `Error formatting argument as str, ignoring message and returning ''` which reflects the function name a bit better.

katcp/core.py
```diff
@@ -299,7 +299,23 @@ def format_argument(self, arg):
...
             return str(int(arg))
         else:
-            return str(arg)
+            try:
+                return str(arg)
+            except UnicodeEncodeError:
+                # unicode characters will break the str cast, so
+                # try to encode to ascii and replace the offending characters
+                # with a '?' character
+                logger.exception("Error casting message argument to str!")
```
ludwigschwardt: Should this also be an exception?

katcp/core.py
```diff
@@ -299,7 +299,23 @@ def format_argument(self, arg):
...
+                try:
+                    if not isinstance(arg, unicode):
+                        arg = arg.decode('utf-8')
+                    return arg.encode('ascii', 'replace')
+                except Exception:
+                    # If all else fails, ignore the string completely
+                    logger.exception(
+                        "Error formatting argument as str, "
+                        "ignoring message and returning ''")
+                    return ''
```
ajoubertza: Do we really want to continue if this happens?  Things are really broken in this case, and we need to investigate.  We have no idea what the message handler will do with the empty string argument we're passing back.

ajoubertza:
@fjoubert Any idea why we don't see a Jenkins build result for this branch?

fjoubert:
@ajoubertza there seems to be something funky going on with the jenkins multibranch jobs, we couldn't get katstore to work with the Jenkinsfiles either. I'll be investigating a bit, but I suspect there is a plugin versioning thing that is causing issues.

Tests pass for me though.

katcp/core.py
fjoubert: The reason we continue is so that a sync can finish even when we get this encoding error. If we just raise a hard exception here, we get stuck in an infinite loop trying to sync with the offending device.

katcp/core.py
fjoubert: I think we can get away with just an error log here because we no longer catch `Exception`

katcp/core.py
ajoubertza: The sync problem was with unicode characters, which is handled now, in the "normal" case.  If it can't handle the unicode, i.e. this last catch all exception, then something really unusual is happening, isn't it?  Not syncing may be better than continuing silently.

On the continuous re-syncing issue, what about backing off for longer before retrying if we get an exception during sync?

fjoubert: I think you are right that it will be good enough to only handle the `UnicodeEncodeError` here and leave other exceptions unhandled. We definitely need to somehow spot when a device's messages are that bad. Besides, I'm not even sure how to get to the second except block.

I'm going to leave out digging into re-syncing now because I suspect that is a client issue that we need to investigate when we see it again.

ajoubertza: OK, we can do that.

ajoubertza:
Will wait until Jenkins is happy again before approving this.

katcp/core.py
```diff
@@ -299,7 +299,17 @@ def format_argument(self, arg):
...
+                return str(arg)
+            except UnicodeEncodeError:
+                # unicode characters will break the str cast, so
+                # try to encode to ascii and replace the offending characters
+                # with a '?' character
+                logger.error("Error casting message argument to str! "
+                             "Trying to encode argument to ascii.")
+                if not isinstance(arg, unicode):
+                    arg = arg.decode('utf-8')
+                return arg.encode('ascii', 'replace')
```
lvdheever: Since there are no easy way to identify the message in the error log (it seems) I suggest to replace the arg with '' as it is easier to notice than finding one ? in a long description.

PR #196, opened 2018-01-18
Allow sampling strategy to be removed from cache

ajoubertza:
When using a `KATCPClientResource`, once a sampling strategy has been set it gets cached and there is no way to remove it.  It can be set to 'none', but after disconnect/sync events there
will still be attempts to apply it.  This can cause unnecessary errors and exceptions to be logged when many sensors that were being sampled are removed from the server's KATCP interface.  The
errors occur during `reapply_sampling_strategies`.  It might be possible to fix the errors by changing something in the sync process, but not sure.

The workaround is to provide a mechanism to drop cached sensor strategies on demand.  This is now available as `drop_sampling_strategy`.

JIRA: [CB-2544](https://skaafrica.atlassian.net/browse/CB-2544)

### DIFF
--- a/katcp/test/test_resource_client.py
+++ b/katcp/test/test_resource_client.py
@@ -583,6 +583,57 @@ class test_KATCPClientResource_IntegratedTimewarp(TimewarpAsyncTestCase):
         self.assertEqual(DUT._sensor_strategy_cache['an_int'], ('event',))
 
     @tornado.testing.gen_test(timeout=1000)
+    def test_drop_sampling_strategy(self):
+        self.server.stop()
+        self.server.join()
+        DUT = resource_client.KATCPClientResource(self.default_resource_spec)
+        DUT.start()
+        yield tornado.gen.moment
+
+        # Check no error for a non-existant sensor
+        DUT.drop_sampling_strategy('not_a_valid_sensor')
+
+        test_strategy = ('period', '2.5')
+        yield DUT.set_sampling_strategy('an_int', test_strategy)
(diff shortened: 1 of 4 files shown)


ajoubertza:
@fjoubert That would be a more intuitive solution.  I thought that applying the 'none' strategy might fail the first time and if it is immediately removed from the cache it won't be retried.  Could we end up with an old sampling strategy remaining active even though you requested it change to 'none'? This is just hypothetical - I didn't investigate the likelihood of such a situation occurring.
@lvdheever Any thoughts on this?

katcp/test/test_resource_client.py
```diff
@@ -582,6 +582,57 @@ def test_set_sampling_strategies(self):
...
+        # Drop the strategy from the cache
+        DUT.drop_sampling_strategy('an_int')
+        self.assertNotIn('an_int', DUT._sensor_strategy_cache)
+        self.assertNotIn('an.int', DUT._sensor_manager._strategy_cache)
+
+        # restart server and check that strategy was NOT applied
+        self.server.start()
+        yield DUT.until_synced()
+        self.assertNotEqual(DUT.sensor.an_int.sampling_strategy, test_strategy)
+
```
lvdheever: Nice comprehensive test.

lvdheever:
@ajoubertza and I had a chat about clearing the cache on strategy = 'None' and although it would be better we decided against that because it may introduce subtle behaviour differences in our system's communication layer that we are not currently aware of that we rely upon. And we do not have the time now and it is quite difficult to get confidence on that level before deploying to site. So we will leave None strategy in the cache for now. A user that wants to remove it can use the new functionality.

PR #197, opened 2018-03-22
Add bulk sensor sampling feature

ajoubertza:
Setting the sampling strategies on KATCP devices with large numbers of sensors is slow.  E.g. a device with 10k sensors can take more than a minute, doing it one sensor at time.  Setting the strategies in bulk requires less roundtrips, making it quicker.

This feature is not enabled by default.  Details are in the [Rev 5.1 specification](https://github.com/ska-sa/katcp-python/blob/master/doc/NRF-KAT7-6.0-IFCE-002-Rev5-1.pdf).

JIRA: [CB-2748](https://skaafrica.atlassian.net/browse/CB-2748)


katcp/server.py
```diff
@@ -2318,6 +2338,99 @@ def inform_callback(sensor, reading):
...
+                        params = [float(params[0]) * MS_TO_SEC_FAC] + params[1:]
+                    requested_strategy = SampleStrategy.get_strategy(
+                        strategy, inform_callback, sensor, *params, ioloop=self.ioloop)
+                    if requested_strategy:
+                        break
+            except Exception:
+                pass
+
+        if not requested_strategy:
+            raise FailReply("No valid strategy provided: %s." % msg.arguments)
```
bmerry: Won't this break if someone tries to query multiple sensors (which the request docstring seems to imply is valid)?

katcp/server.py
```diff
@@ -2224,10 +2232,10 @@ def request_sensor_sampling(self, req, msg):
         -------
         success : {'ok', 'fail'}
             Whether the sensor-sampling request succeeded.
-        name : str
-            Name of the sensor queried or configured.
+        names : list str
+            Name(s) of the sensor queried or configured.
         strategy : {'none', 'auto', 'event', 'differential', 'period'}
-            Name of the new or current sampling strategy for the sensor.
+            Name(s) of the new or current sampling strategy for the sensor(s).
```
lvdheever: "strategy" parameter is missing "differential-rate" from before

katcp/server.py
```diff
@@ -2241,10 +2249,22 @@ def request_sensor_sampling(self, req, msg):
             ?sensor-sampling cpu.power.on period 500
             !sensor-sampling ok cpu.power.on period 500
 
+            if BULK_SET_SENSOR_SAMPLING is enabled then:
+
+            ?sensor-sampling cpu.power.on fan.speed
+            !sensor-sampling fail Cannot\_query\_multiple\_sensors
+
+            ?sensor-sampling cpu.power.on fan.speed period 500
+            !sensor-sampling ok cpu.power.on fan.speed period 500
```
lvdheever: Would have been nicer to have the name or list of names at the end - but unfortunately legacy dictates otherwise...

katcp/server.py
```diff
@@ -2318,6 +2338,99 @@ def inform_callback(sensor, reading):
...
+                        strategy, inform_callback, sensor, *params, ioloop=self.ioloop)
+                    if requested_strategy:
+                        break
+            except Exception:
+                pass
+
+        if not requested_strategy:
+            raise FailReply("No valid strategy provided: %s." % msg.arguments)
+
+        # with known strategy, remaining arguments are sensor names
```
lvdheever: May be change comment for clarity "preceding arguments are sensor names" instead of "remaining"
(or it could be a slightly incorrect Afrikaans definition of "remaining")

katcp/server.py
```diff
@@ -2318,6 +2338,99 @@ def inform_callback(sensor, reading):
...
+        # Let the ioloop run so that the #sensor-status informs are sent before
+        # the reply. Not strictly neccesary, but a number of tests depend on
+        # this behaviour, less effort to fix it here :-/
+        yield gen.moment
+        reply_args = names + [strategy] + params
+        raise gen.Return(req.make_reply("ok", *reply_args))
+
     @request()
     @return_reply()
     def request_sensor_sampling_clear(self, req):
```
lvdheever: Not sure if there is enough of a use case to also update sensor_sampling_clear to support bulk sensors, but that can be done in subsequent PR if required.

katcp/server.py
```diff
@@ -2318,6 +2338,99 @@ def inform_callback(sensor, reading):
...
+        def inform_callback(sensor, reading):
+            """Inform callback for sensor strategy."""
+            timestamp, status, value = reading
+            cb_msg = format_inform(sensor, timestamp, status, value)
+            client.inform(cb_msg)
+
+        # Parse arguments from the back, trying all possible strategies until
+        # a valid one is found.  The number of arguments to define a strategy
+        # is one for the name, plus the number of params expected for that type
+        # of strategy.
```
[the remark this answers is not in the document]
lvdheever: Unfortunately a space is the katcp delimiter, but a comma might do if we update the KATCP spec for 5.1 to exclude comma from sensor-names, as in practice I don't think we have any sensors with comma's in the name.

The katcp spec has the following for the message BNF:
```
<message> ::= <type> <name> [<msgid>] <arguments> <eol>
<type> ::= "?" | "!" | "#"
<name> ::= alpha (alpha | digit | "-")*
<msgid> ::= "[" (digit / "0") digit* "]"
<whitespace> ::= (space | tab) [<whitespace>]
<eol> ::= newline | carriage-return
<arguments> ::= (<whitespace> <argument> <arguments>) | <whitespace> | ""
<argument> ::= (<plain> | <escape>)+
<escape> ::= "\" <escapecode>
<escapecode> ::= "\" | "_" | zero | "n" | "r" | "e" | "t" | "@"
<special> ::= backslash | space | null | newline | carriage-return | escape | tab
<plain> ::= character / <special>
```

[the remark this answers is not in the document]
lvdheever: It just defines "character" (which I don't think is defined in standard BNF) without describing it further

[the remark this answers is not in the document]
bmerry: You can still include a space in an argument, it just has to be escaped. So on the wire it would look something like
```
?sensor-sampling sensor-1\_sensor-2\_sensor-3 period 500
```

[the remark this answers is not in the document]
lvdheever: Yes true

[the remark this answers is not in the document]
ajoubertza: I agree that having to try all the strategies backwards isn't a great solution, and does prevent variable arguments in future.  Passing all the names as the first parameters with a separator character sounds like a much a better idea.  I prefer a comma to an escaped space.  It is much more readable and easier to type in debug session.  It seems almost as unlikely as a space in any existing sensor names. 
```
?sensor-sampling sensor-1,sensor-2,sensor-3 period 500
```
Are commas OK?
Question 2:  should we update the spec's section on sensors to include a list of allowed characters, and if so would that require an increment in the major version number?

katcp/server.py
ajoubertza: It will break, but the request's docstring does say that queries of multiple sensors aren't allowed. 

https://github.com/ska-sa/katcp-python/blob/c5298956cdf3ba25f4240804307a09e0bb287869/katcp/server.py#L2203

Also the example in that docstring shows a 'fail' if multiple sensors are queried, although the error message will differ slightly:
https://github.com/ska-sa/katcp-python/blob/c5298956cdf3ba25f4240804307a09e0bb287869/katcp/server.py#L2254-L2255

katcp/server.py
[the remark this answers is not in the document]
bmerry: I think commas are probably okay. I think the spec generally needs more written about acceptable character sets, but I see that as a clarification rather than a major change - putting weird characters in sensor names is probably already going to break something somewhere.

katcp/server.py
ajoubertza: Thanks.  Using "preceding" would be better.

katcp/server.py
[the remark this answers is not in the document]
ajoubertza: Thanks, then I will go with commas.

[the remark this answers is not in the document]
ajoubertza: Does anyone know where the LaTeX source files are for the KATCP guidelines doc?  Maybe @brickZA's memory goes back to 2012, when they were last updated?!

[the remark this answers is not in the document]
lvdheever: It lives in https://katfs.kat.ac.za/svnSystemEngineering/General/KATCP/NRF-KAT7-6.0-IFCE-002%20(GuidelinesForCommunicationWithDevices)
I think "trunk" has the master source and "tags" the issued versions.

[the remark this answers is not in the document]
lvdheever: I agree with @bmerry that we can handle the update as a clarification of valid characters in sensor names as a minor version change, not major.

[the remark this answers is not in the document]
ajoubertza: @bmerry and @lvdheever Proposed updates to the guidelines are here:  ska-sa/katcp-guidelines/pull/1
Please have a look.

katcp/sampling.py
```diff
@@ -92,6 +92,17 @@ class SampleStrategy(object):
...
+    ## @brief Mapping from strategy constant to number of parameters.
+    SAMPLING_NUM_PARAMS_LOOKUP = {
+        NONE: 0,
+        AUTO: 0,
+        PERIOD: 1,
+        EVENT: 0,
+        DIFFERENTIAL: 1,
+        EVENT_RATE: 2,
+        DIFFERENTIAL_RATE: 3,
+    }
```
mmphego, line 104: We should probably fix this, since we are delimiting by `,` instead of ` `

ajoubertza: This dict is no longer required, and has been removed.

katcp/server.py
```diff
@@ -2237,13 +2250,25 @@ def request_sensor_sampling(self, req, msg):
...
+            ?sensor-sampling cpu.power.on period 0.5
+            !sensor-sampling ok cpu.power.on period 0.5
+
+            if BULK_SET_SENSOR_SAMPLING is enabled then:
+
+            ?sensor-sampling cpu.power.on,fan.speed
+            !sensor-sampling fail Cannot\_query\_multiple\_sensors
+
+            ?sensor-sampling cpu.power.on,fan.speed period 0.5
+            !sensor-sampling ok cpu.power.on,fan.speed period 0.5
```
mmphego, line 2262: Perhaps, update the docstring to match the actual implementation

katcp/server.py
```diff
@@ -2237,13 +2250,25 @@ def request_sensor_sampling(self, req, msg):
...
 
         """
+        if self.PROTOCOL_INFO.bulk_set_sensor_sampling and len(msg.arguments) > 1:
+            handler = self._handle_bulk_set_sensor_sampling
+        else:
+            handler = self._handle_sensor_sampling
```
mmphego, line 2268: This implementation will fail when one requests multiple sensors without passing a strategy.
eg: 
```bash
> ?sensor-sampling bandwidth,n-chans
< !sensor-sampling fail Unknown sensor name: bandwidth,n-chans. 
```

This should reply with the last known value or strategy for the sensors passed.

mmphego (changes requested):

katcp/test/test_server.py
```diff
@@ -462,6 +541,119 @@ def test_request_timeout_hint(self):
...
+        self.server.request_sensor_sampling(req, req.msg).result()
+        inform = req.client_connection.inform
+        inform.assert_wait_call_count(count=2)
+        inform_messages = [call.args[0] for call in inform.call_args_list]
+        self._assert_msgs_equal(
+            inform_messages, (
+                r"#sensor-status 12345.000000 1 an.int nominal 1.5",
+                r"#sensor-status 12345.000000 1 a.float nominal 1.5",
+            )
+        )
```
mmphego, line 590: Currently fails, we need to add the implementation in the code.

katcp/server.py
```diff
@@ -2318,6 +2343,74 @@ def inform_callback(sensor, reading):
...
+        params = msg.arguments[2:]
+
+        invalid_names = []
+        for name in names:
+            if name not in self._sensors:
+                invalid_names.append(name)
+        if invalid_names:
+            raise FailReply("Unknown sensor names: %r." % invalid_names)
+
+        requested_strategies = {}
```
bngcebetsha, line 2372: We should make it clear we are only doing one strategy per bulk sensor sampling

katcp/server.py
ajoubertza: This request should fail, according to the guidelines you cannot query the strategies for multiple sensors.  But we could improve the error message, if bulk sensor sampling is enabled.

katcp/test/test_server.py
mmphego: See https://github.com/ska-sa/katcp-python/pull/197#discussion_r513340242

katcp/server.py
```diff
@@ -2189,19 +2189,28 @@ def request_sensor_value(self, req, msg):
         return req.make_reply("ok", str(len(sensors)))
 
     def request_sensor_sampling(self, req, msg):
-        """Configure or query the way a sensor is sampled.
+        r"""Configure or query the way a sensor is sampled.
```
lanceWilliams, line 2192: Typo?

bxaia: I think the r is for the code to treat that sentence as raw string

katcp/test/test_server.py
```diff
@@ -333,6 +333,84 @@ def test_on_client_connect(self):
...
+            3.5,
+            client_conn=client
+        )
+        reply = self.server.request_sensor_sampling(req, req.msg).result()
+        self._assert_msgs_equal(
+            [reply], ["!sensor-sampling ok a-sens differential-rate 1.5 2.5 3.5"]
+        )
+
+    def test_bulk_sensor_sampling_fails_if_not_supported(self):
+        if not self.server.PROTOCOL_INFO.bulk_set_sensor_sampling:
```
lanceWilliams, line 394: This doesn't seem to test the case where bulk sensor sampling is disabled, this if statement would just skip everything below since the test server is started with this flag set.

katcp/test/test_server.py
```diff
@@ -333,6 +333,84 @@ def test_on_client_connect(self):
...
+            start_thread_with_cleanup(self, self.server)
+            self.server._send_message = WaitingMock()
+            self.server.wait_running(timeout=1.0)
+            self.assertTrue(self.server.running())
+
+            req = mock_req("sensor-sampling", "an.int,a.float", "period", 1.5)
+            with self.assertRaises(FailReply):
+                self.server.request_sensor_sampling(req, req.msg).result()
+
+    def test_querying_multiple_sensor_sampling_strategies_fails(self):
```
lanceWilliams, line 404: This test should maybe be moved to test_DeviceServer51 since that has the server with the bulk flag set. It think this test here would raise fail reply for the flag not being set as opposed to the contents of the request which is what we want.

katcp/test/test_server.py
lanceWilliams, line 394: If the flag is not set for the test server in this class, why do we do this if statement?

katcp/server.py
ajoubertza: Yes, it keeps the linter happy, because we have escaped strings, like `\_` within the docstring.

katcp/test/test_server.py
ajoubertza: Good point.  I think we should delete this one, since we have the same test already in the v51  class:  `test_bulk_sensor_sampling_fails_for_no_strategy`.

katcp/test/test_server.py
ajoubertza: We do this because the test also runs in the v51 tests (which inherit from `test_DeviceServer`).  The test isn't relevant for the v51 server which has bulk sampling enabled.

katcp/core.py
```diff
@@ -660,6 +661,9 @@ class ProtocolFlags(object):
     # New proposal flag to indicate that a device supports ?request-timeout-hint
     # See CB-2051
     REQUEST_TIMEOUT_HINTS = b'T'
+    # New proposal flag to indicate that a device supports configuring multiple
+    # sensors in a single ?sensor-sampling request
+    BULK_SET_SENSOR_SAMPLING = b'B'
```
ajoubertza, line 666: These comments should be removed - the flags are no longer "new proposals".  They have been accepted.

katcp/sampling.py
```diff
@@ -530,6 +530,10 @@ def __init__(self, inform_callback, sensor, *params, **kwargs):
             raise ValueError("The 'differential-rate' strategy takes three parameters.")
         super(SampleDifferentialRate, self).__init__(inform_callback, sensor,
                                                      *params[1:], **kwargs)
+        # Fix up the parameters so we do not loose the params that are not
```
ajoubertza, line 533: Typo
```suggestion
        # Fix up the parameters so we do not lose the params that are not
```

katcp/server.py
```diff
@@ -2217,16 +2226,20 @@ def request_sensor_sampling(self, req, msg):
...
-        name : str
-            Name of the sensor queried or configured.
-        strategy : {'none', 'auto', 'event', 'differential', 'period'}
-            Name of the new or current sampling strategy for the sensor.
+        names : str
+            Name(s) of the sensor queried or configured. If multiple sensors, this will
+            be a comma-separated list.
+        strategy : {'none', 'auto', 'event', 'differential', 'differential-rate',
+                    'period', 'event-rate'}.
+            Name(s) of the new or current sampling strategy for the sensor(s).
```
ajoubertza, line 2242: Typo.  There is only one strategy that will be returned.
```suggestion
            Name of the new or current sampling strategy for the sensor(s).
```

katcp/server.py
ajoubertza: This does match the request and reply, but maybe you mean the informs should be included?  We could add that, and then the same for lines 2253-2254.

katcp/server.py
```diff
@@ -2318,6 +2351,79 @@ def inform_callback(sensor, reading):
...
+        def inform_callback(sensor, reading):
+            """Inform callback for sensor strategy."""
+            timestamp, status, value = reading
+            cb_msg = format_inform_v5(sensor, timestamp, status, value)
+            client.inform(cb_msg)
+
+        names_arg = ensure_native_str(msg.arguments[0])
+        names = names_arg.split(',')
+
+        if not len(msg.arguments) > 1:
```
ajoubertza, line 2371: Would it be easier to read if removed the `not`?
```suggestion
        if len(msg.arguments) <= 1:
```

katcp/test/test_server.py
```diff
@@ -462,6 +541,127 @@ def test_request_timeout_hint(self):
...
+        with self.assertRaises(FailReply):
+            self.server.request_sensor_sampling(req, req.msg).result()
+
+    def test_bulk_sensor_sampling_fails_for_no_strategy(self):
+        client = self.set_up_client_for_bulk_sensor_sampling()
+
+        req = mock_req("sensor-sampling", "an.int,a.float", client_conn=client)
+        with self.assertRaises(FailReply):
+            self.server.request_sensor_sampling(req, req.msg).result()
+
```
ajoubertza, line 664: One more test has come to mind.  The spec says that the implementation must be atomic.  I.e, if we can't set the strategies on all the sensors, then there must be no change to any of them.  Something like: 
`test_bulk_sensor_sampling_all_strategies_unchanged_for_failure`.

ajoubertza (commented):
A few minor changes, please.

ajoubertza (commented):
LGTM
(although GitHub won't let me approve my own PR!)

PR #206, opened 2019-07-12
Fix some memory leaks

ajoubertza:
There are some memory leaks in KATCP's inspecting clients, resource clients and containers, and server.  This PR is to fix the known issues.

### Issue 1
As background, existing client functionality allows it to either create its own ioloop instance on a new thread (managed), or allow the caller to provide the ioloop (unmanaged).  This choice typically determines the thread that makes the call to the `stop()` method.  The unmanaged case will normally use the same thread as the ioloop, while the managed case will normally be stopped from a different thread.

If the `DeviceClient` doesn't exit its main coroutine after being stopped, a reference to the client object remains, and it cannot be cleaned up by the garbage collector.  This can result
in a memory leak.  In the "managed" case, the stop method wasn't giving the main coroutine enough time to exit.  In the the "unmanaged" case, if the user keeps calling `yield`, the object will eventually be collectable. 

The client's stop method now works properly for both cases.  When using the same thread as the ioloop (unmanaged), the caller should yield on the new `until_stopped()` method rather than calling `join()`.  When stopping from a different thread (managed), then the opposite applies - use `join()`, not `until_stopped()`.

It is common to use the inspecting client and resource clients in the unmanaged ioloop mode, so it is better if they wait using the new `until_stopped()` method.

Added unit tests to verify the objects can be freed.  They use the simple approach of checking that a weak reference is cleared.

### Issue 2
If a user is waiting on the disconnected state of a resource and then that resource is stopped, the future will never resolve.  For example, when using the `monitor_resource_sync_state` function, which watches a container's state.  This future has a reference to the container, so the container cannot be released by the garbage collector.  This can result in a memory leak, noticeable in long running applications that create and destroy client resource containers and watch them with the `monitor_resource_sync_state` method (e.g. CAM's subarray manager).   We now send a state update when stopping the inspecting client that indicates that it is disconnected.  This state update allows futures waiting on `until_not_synced` to complete.

A couple of related tests were added.  There is still a possible memory leak if stopping a resource container while it is syncing, but a fix hasn't been found yet.  A test was added - it is marked
to be skipped until such time as the problem is fixed.

JIRA: [MT-834](https://skaafrica.atlassian.net/browse/MT-834)

### DIFF
--- a/katcp/client.py
+++ b/katcp/client.py
@@ -886,30 +886,21 @@ class DeviceClient(object):
         ioloop = getattr(self, 'ioloop', None)
         if not ioloop:
             raise RuntimeError('Call start() before stop()')
+        stopped_from_ioloop_thread = get_thread_ident() == self.ioloop_thread_id
 
-        def _stop_from_non_ioloop_thread():
+        @tornado.gen.coroutine
+        def _stop():
             if timeout:
-                self._running.wait_with_ioloop(self.ioloop, timeout)
+                yield self._running.until_set(timeout)
             # clear _running before disconecting to allow line_read_loop
             # and connect_loop to exit
             self._running.clear()
             self._disconnect()
-            self._stopped.wait_with_ioloop(self.ioloop, timeout)
+            if not stopped_from_ioloop_thread:
+                yield self._stopped.until_set(timeout)
+            # else unmanaged ioloop, so caller should yield on until_stopped()
 
-        @tornado.gen.coroutine
-        def _stop_from_ioloop_thread():
-            if timeout:
-                yield self._running.until_set(timeout)
-            self._running.clear()
-            self._disconnect()
-            # Note:  caller should yield on until_stopped()
-
-        if get_thread_ident() == self.ioloop_thread_id:
-            callback = _stop_from_ioloop_thread
-        else:
-            _stop_from_non_ioloop_thread()
-            callback = None
-        self._ioloop_manager.stop(timeout=timeout, callback=callback)
+        self._ioloop_manager.stop(timeout=timeout, callback=_stop)
 
     def until_stopped(self, timeout=None):
         """Return future that resolves when the client has stopped.
--- a/katcp/client.py
+++ b/katcp/client.py
@@ -147,6 +147,9 @@ class DeviceClient(object):
         self._bindaddr = (host, port)
         self._tb_limit = tb_limit
         self._logger = logger
+        # Indicate that client's main coroutine is not running
+        self._stopped = AsyncEvent()
+        self._stopped.set()
         # Indicate that client is running and ready connect
         self._running = AsyncEvent()
         # Indicate that the client is waiting for a reconnect. Purely for testing.
@@ -639,36 +642,41 @@ class DeviceClient(object):
 
     @gen.coroutine
     def _install(self):
-        ioloop_before = self.ioloop
-        # Do stuff to put us on the IOLoop
-        self._logger.debug("Starting client loop for {0!r}"
-                           .format(self._bindaddr))
-        self.ioloop_thread_id = get_thread_ident()
-        self._tcp_client = tornado.tcpclient.TCPClient()
-        self._running.set()
-        yield self._connect()
-        if self._stream is None and not self._auto_reconnect:
-            raise KatcpClientError("Failed to connect to {0!r}"
-                                   .format(self._bindaddr))
+        self._stopped.clear()
         try:
-            yield self._connect_loop()
-        except Exception:
-            self._logger.exception('Unhandled exception in _connect_loop()')
-        finally:
+            ioloop_before = self.ioloop
+            # Do stuff to put us on the IOLoop
+            self._logger.debug("Starting client loop for {0!r}"
+                               .format(self._bindaddr))
+            self.ioloop_thread_id = get_thread_ident()
+            self._tcp_client = tornado.tcpclient.TCPClient()
+            self._running.set()
+            yield self._connect()
+            if self._stream is None and not self._auto_reconnect:
+                raise KatcpClientError("Failed to connect to {0!r}"
+                                       .format(self._bindaddr))
             try:
-                # Make sure everything is torn down properly
-                if ioloop_before == self.ioloop:
-                    # But only if we are still using the same ioloop. If the ioloop has
-                    # changed that means this client has been stopped and re-started
-                    # before we could get back to this finally clause. This would result
-                    # in us stopping the new ioloop. D'oh!
-                    self.ioloop.add_callback(self.stop)
-            except RuntimeError, e:
-                if str(e) == 'IOLoop is closing':
-                    # Seems the ioloop was stopped already, no worries.
-                    self._running.clear()
-            self._logger.info("Client connect loop for {0!r} finished."
-                              .format(self._bindaddr))
+                yield self._connect_loop()
+            except Exception:
+                self._logger.exception('Unhandled exception in _connect_loop()')
+            finally:
+                try:
+                    # Make sure everything is torn down properly
+                    if ioloop_before == self.ioloop:
+                        # But only if we are still using the same ioloop. If the ioloop has
+                        # changed that means this client has been stopped and re-started
+                        # before we could get back to this finally clause. This would result
+                        # in us stopping the new ioloop. D'oh!
+                        if self.running():
+                            self.stop()
+                except RuntimeError, e:
+                    if str(e) == 'IOLoop is closing':
+                        # Seems the ioloop was stopped already, no worries.
+                        self._running.clear()
+                self._logger.info("Client connect loop for {0!r} finished."
+                                  .format(self._bindaddr))
+        finally:
+            self._stopped.set()
 
     @gen.coroutine
     def _connect_loop(self):
@@ -682,7 +690,9 @@ class DeviceClient(object):
                 self.ioloop.add_callback(self._waiting_to_retry.set)
                 yield until_later(self.auto_reconnect_delay)
                 self._waiting_to_retry.clear()
-                yield self._connect()
+                # client may have been stopped while waiting on reconnect delay
+                if self.running():
+                    yield self._connect()
             else:
                 break
             # Allow stream-close handlers etc to run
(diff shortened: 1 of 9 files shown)


katcp/testutils.py
```diff
@@ -2049,3 +2051,41 @@ def _advance(self, future):
         yield self.test_instance.set_ioloop_time(
             self.test_instance.ioloop_time + self.quantum)
         future.set_result(None)
+
+
+@contextmanager
+def assert_no_memory_leaks():
+    """Context handler that asserts no memory is leaking within.
```
ajoubertza: @bmerry PR still in draft, but can you please have a look at this approach to checking for memory leaks?  It seems to pick up the problems.

Suggest first looking at this context handler, and then [its tests](https://github.com/ska-sa/katcp-python/pull/206/files#diff-b7551ec24900e8af18e8858ea2d00622R208).

katcp/testutils.py
```diff
@@ -2049,3 +2051,41 @@ def _advance(self, future):
...
+        # check for initial garbage (e.g., test framework may have other leaks)
+        gc.collect()
+        initial_garbage = set(id(g) for g in gc.garbage)
+        yield
+        # collect again and see if there is any new garbage
+        gc.collect()
+        final_garbage = set(id(g) for g in gc.garbage)
+        new_garbage = final_garbage.difference(initial_garbage)
+        unreachable = len(new_garbage)
+        assert unreachable == 0, "{} unreachable objects found".format(unreachable)
```
bmerry: Using `assert` rather than the unittest assertions means that the assert will disappear if run with `python -O` (unless. nose.tools provides the unittest assertions outside of the TestCase class e.g. `nose.tools.assert_equal(0, unreachable)`.

katcp/testutils.py
```diff
@@ -2049,3 +2051,41 @@ def _advance(self, future):
...
+
+    Raises
+    ------
+    AssertionError
+        If the garbage collector finds any unreachable objects when
+        exiting the context handler.
+
+    """
+    old_config = gc.get_debug()
+    gc.set_debug(gc.DEBUG_SAVEALL)
```
bmerry: Are you sure you want this? It means that the test will fail if there happens to be cyclic garbage left over that *can* be garbage-collected (just not as quickly as the reference-counter would free it), as opposed to only failing on uncollectable garbage.

If this is what you want, I would suggest:
- Setting the new flags to `old_config | gc.DEBUG_SAVEALL` so that any other flags set for debugging aren't wiped.
- Doing the initial `gc.collect()` before setting the flag, so that cyclic but collectable garbage that existed before the test can be freed rather than being stuck forever in `gc.garbage`.

katcp/test/test_testutils.py
```diff
@@ -202,3 +203,26 @@ def test_ioloop_hogging(self):
...
+class test_AssertNoMemoryLeaks(unittest.TestCase):
+
+    def test_non_leaky_object(self):
+
+        class NonLeaky(object):
+            pass
+
+        with testutils.assert_no_memory_leaks():
+            obj = NonLeaky()
+            obj = None  # noqa: F841
```
bmerry: You could replace both lines with just `NonLeaky()`.

katcp/test/test_testutils.py
```diff
@@ -202,3 +203,26 @@ def test_ioloop_hogging(self):
...
+    def test_leaky_object(self):
+
+        class Leaky(dict):
+            def __init__(self):
+                self.__dict__ = self
+
+        with self.assertRaises(AssertionError):
+            with testutils.assert_no_memory_leaks():
+                obj = Leaky()
+                obj = None  # noqa: F841
```
bmerry: It might be a good idea to have an object that has cyclic references but is collectable e.g.
```python
class Cyclic(object):
    def __init__(self):
        self.me = self
```
and test it one way or the other depending on whether you want to allow this.

katcp/testutils.py
mmphego: @bmerry sorry to sidetrack you, but when would you need to run `python -O`?

bmerry: You might want to verify that the system under test functions correctly with `-O`. I vaguely recall that some project I contributed to (possibly dask) does this as part of CI because there had been such a bug at one point.

It could also happen accidentally if someone sets `PYTHONOPTIMIZE` in their environment in an attempt to speed up all their Python programs.

katcp/testutils.py
```diff
@@ -2049,3 +2051,41 @@ def _advance(self, future):
...
+    try:
+        # check for initial garbage (e.g., test framework may have other leaks)
+        gc.collect()
+        initial_garbage = set(id(g) for g in gc.garbage)
+        yield
+        # collect again and see if there is any new garbage
+        gc.collect()
+        final_garbage = set(id(g) for g in gc.garbage)
+        new_garbage = final_garbage.difference(initial_garbage)
```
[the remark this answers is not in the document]
ajoubertza: Thanks, will do.

katcp/testutils.py
ajoubertza: Interesting that you mention that - one of my earlier iterations did pass in the test case object, so I can go back to that.  I thought using a vanilla `assert` would be more generally useful, but the only place we really want to use this context manager is in our own tests.

katcp/test/test_testutils.py
ajoubertza: Doh!  That's a better idea.

katcp/test/test_testutils.py
ajoubertza: This one isn't collectable (at least not in Python 2.7 and 3.5).  Should it to be?

bmerry: It seems collectable to me, in both 2.7 and 3.6:
```python3
import gc

class Cyclic(object):
    def __init__(self):
        self.me = self


Cyclic()
gc.collect()
print(gc.garbage)
```

prints an empty list.

ajoubertza: @bmerry OK - I was using the `gc.DEBUG_SAVEALL` flag, but we shouldn't.  The objects are being freed.  Back to the drawing board.

katcp/test/test_resource_client.py
```diff
@@ -1,4 +1,3 @@
-from __future__ import print_function, division, absolute_import
```
ajoubertza: Removed a duplicate - already on line 9

ajoubertza:
@bmerry I've reverted the first commit that was testing for memory leaks incorrectly.  New changes, e9958bd, fix one of the major issues, and test in a better way.  I've put the details in the updated PR description.

There is a second issue I'm working on, but seems like it will be a much smaller change, so this first part can be reviewed.

bmerry:
Ok I can take a look. Should I look at the whole thing or just the tests?

katcp/client.py
```diff
@@ -639,36 +642,41 @@ def unhandled_request(self, msg):
...
+            except Exception:
+                self._logger.exception('Unhandled exception in _connect_loop()')
+            finally:
+                try:
+                    # Make sure everything is torn down properly
+                    if ioloop_before == self.ioloop:
+                        # But only if we are still using the same ioloop. If the ioloop has
+                        # changed that means this client has been stopped and re-started
+                        # before we could get back to this finally clause. This would result
+                        # in us stopping the new ioloop. D'oh!
```
bmerry: In that case (where we've restarted before getting here), we're still going to clear `_stopped` in the final `finally` clause. Is that a bug?

ajoubertza:
> Should I look at the whole thing or just the tests?

Please look at `test_client.py` and `client.py` for starters.

> Also I'll probably only get to this some time next week

That's fine.  Thanks.

katcp/client.py
ajoubertza: Good question.  Think it is a bug.  If the user has changed the ioloop, but hasn't called `start` a second time then we would be OK.  However, if they have called `start` again, the behaviour would be incorrect.

Currently the `start` method fails if `_running` is already set.  If I include a check there for the new `_stopped` flag as well, then we should be safe.

katcp/client.py
```diff
@@ -852,32 +862,81 @@ def join(self, timeout=None):
 
         Notes
         -----
-        Does nothing if the ioloop is not managed.
+        Does nothing if the ioloop is not managed.  Use until_stopped() instead.
```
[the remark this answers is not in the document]
ajoubertza: Intended.

katcp/client.py
```diff
@@ -639,36 +642,41 @@ def unhandled_request(self, msg):
...
+                try:
+                    # Make sure everything is torn down properly
+                    if ioloop_before == self.ioloop:
+                        # But only if we are still using the same ioloop. If the ioloop has
+                        # changed that means this client has been stopped and re-started
+                        # before we could get back to this finally clause. This would result
+                        # in us stopping the new ioloop. D'oh!
+                        if self.running():
+                            self.stop()
+                except RuntimeError, e:
```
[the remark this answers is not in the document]
ajoubertza: True.  Most of katcp still uses the old syntax, so I was sticking to that.

katcp/client.py
ajoubertza: I tried to test this edge case, but doesn't seem like we could ever hit it:
- if the ioloop is unmanaged it cannot be changed to a new instance, even if we have stopped the client.  [This error](https://github.com/ska-sa/katcp-python/blob/232647424019c72c9b384e9bcfaad5f075009871/katcp/ioloop_manager.py#L59) is raised.
- if the ioloop is managed, stopping and starting again makes a new ioloop instance.  If you don't call `join` after `stop`, the next call to `start` [fails](https://github.com/ska-sa/katcp-python/blob/232647424019c72c9b384e9bcfaad5f075009871/katcp/ioloop_manager.py#L83) as the old ioloop's thread is still running.  If you do call `join`, the cleanup will wait for `_install` to be done, in which case we don't hit the problem. 
- if the ioloop was initially managed, we can change it to a new unmanaged ioloop instance, but only after calling `stop` and `join`.

Conclusion:  no need to add the check for `_stopped` in `start`.  I have added another test.

bmerry: Does that also mean that this `if` statement isn't needed?

ajoubertza: Yes.  Although an ill-advised user could modify `self.ioloop` directly.  It would be a little safer if it was changed to a read-only property.

katcp/client.py
```diff
@@ -852,32 +860,81 @@ def join(self, timeout=None):
...
+        If using an unmanaged ioloop, this can be called from the
+        same thread as the ioloop.  The `until_stopped()` method can be
+        used to wait on completion of the main coroutine, `_install()`.
 
         Parameters
         ----------
         timeout : float in seconds
-           Seconds to wait for client thread to have *started*.
+           Seconds to wait for both client thread to have *started*, and
+           for stopping.
```
bmerry: What happens if the client thread doesn't stop in time? Is an exception raised?

bmerry:
I'm not entirely following where the memory leak comes from. You first say "If the DeviceClient's ioloop is stopped before the main coroutine has completed", but in the unmanaged case you presumably have just one ioloop for the lifetime of the program which presumably won't be stopped until shutdown?

katcp/client.py
```diff
@@ -852,32 +860,81 @@ def join(self, timeout=None):
...
+            self._running.clear()
+            self._disconnect()
+            # Note:  caller should yield on until_stopped()
+
+        if get_thread_ident() == self.ioloop_thread_id:
+            callback = _stop_from_ioloop_thread
+        else:
+            _stop_from_non_ioloop_thread()
+            callback = None
+        res = self._ioloop_manager.stop(timeout=timeout, callback=callback)
```
bmerry: `res` doesn't seem to be used anywhere.

katcp/test/test_client.py
```diff
@@ -8,13 +8,14 @@
...
-import mock
-import time
+import gc
 import logging
+import mock
 import threading
+import time
 import tornado.testing
+import unittest2 as unittest
+import weakref
```
bmerry: PEP8 recommends putting stdlib modules before 3rd-party modules

katcp/test/test_client.py
```diff
@@ -310,6 +315,182 @@ def notify_connected(connected):
...
+    def test_no_memory_leak_change_ioloop(self):
+        client = katcp.DeviceClient(self.host, self.port)
+        wr = weakref.ref(client)
+
+        # start and stop client with managed ioloop
+        client.start(timeout=0.1)
+        client.wait_protocol(timeout=0.1)
+        self.assertTrue(client.protocol_flags)
+        client.stop(timeout=0.1)
+        client.join(timeout=0.1)
```
bmerry: The start/wait_protocol/assert/stop/join block is copy-pasted a lot, and could go into a helper function. A similar helper could cover the unmanaged case.

katcp/test/test_client.py
```diff
@@ -310,6 +315,182 @@ def notify_connected(connected):
...
+
+        client.start(timeout=0.1)
+        client.wait_protocol(timeout=0.1)
+        self.assertTrue(client.protocol_flags)
+        client.stop(timeout=0.1)
+        client.join(timeout=0.1)
+
+        # clear strong reference and check if object can be garbage collected
+        client = None
+        gc.collect()
```
bmerry: I don't know if katcp aims to support pypy at all, but my experience has been that pypy sometimes needs multiple calls to `gc.collect()` to fully collect all unreachable objects (it cleans out objects that have no references, but doesn't recurse to objects that those collected objects were referencing).

katcp/client.py
```diff
@@ -749,7 +757,7 @@ def set_ioloop(self, ioloop=None):
 
         """
         self._ioloop_manager.set_ioloop(ioloop, managed=False)
-        self.ioloop = ioloop
+        self._ioloop = ioloop
```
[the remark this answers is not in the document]
ajoubertza: `_ioloop_manager` handles this.  It only allows the `ioloop` to be set once, [raising an error](https://github.com/ska-sa/katcp-python/blob/232647424019c72c9b384e9bcfaad5f075009871/katcp/ioloop_manager.py#L59) on subsequent attempts.

katcp/client.py
```diff
@@ -852,32 +860,81 @@ def join(self, timeout=None):
...
-        """Stop a running client (from another thread).
+        """Stop a running client.
+
+        If using a managed ioloop, this must be called from a different
+        thread to the ioloop's.  This method only returns once the client's
+        main coroutine, `_install()`, has completed.
+
+        If using an unmanaged ioloop, this can be called from the
+        same thread as the ioloop.  The `until_stopped()` method can be
+        used to wait on completion of the main coroutine, `_install()`.
```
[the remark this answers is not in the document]
ajoubertza: We probably could return a future.  I'm also no sure what tornado will do with a result that isn't yielded on.

I agree that having a method sometimes work like a coroutine and sometimes not is confusing.  I didn't want to change the semantics for existing users which is why I added the `until_stopped` method instead.  It is also similar to the other `until_xyz` methods already in use.

katcp/client.py
ajoubertza: Doesn't look like an exception is raised, so caller won't know.  Guess code might get stuck when waiting on `join` or `until_stopped`.  This is how it was working before, so I don't plan on fixing it.

katcp/test/test_client.py
ajoubertza: Interesting.  I've never thought about using katcp in pypy  - no plans to support it.  We can update the tests, if that ever happens.

ajoubertza:
> I'm not entirely following where the memory leak comes from. You first say "If the DeviceClient's ioloop is stopped before the main coroutine has completed", but in the unmanaged case you presumably have just one ioloop for the lifetime of the program which presumably won't be stopped until shutdown?

Oops - I shouldn't have said ioloop there.  Maybe this is better:  "If the `DeviceClient` doesn't exit its main coroutine after being stopped...".  I'll update the description.

katcp/client.py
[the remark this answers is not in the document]
bmerry: Agreed, let's leave it as is.

bmerry:
>  Maybe this is better: "If the DeviceClient doesn't exit its main coroutine after being stopped..."

That makes more sense, but I'm still a little confused why the main coroutine wouldn't exit (after some small amount of time). Is it because `stop` wasn't properly interrupting it?

ajoubertza:
> That makes more sense, but I'm still a little confused why the main coroutine wouldn't exit (after some small amount of time). Is it because `stop` wasn't properly interrupting it?

As discussed, you are correct that the main coroutine will exit after a while (5 to 10 more `yield`s) in the unmanaged case.  In the managed case, the ioloop was being stopped too soon.  The memory leak in unmanaged case was due to issue 2, in the description.

PR #210, opened 2019-08-23
Use correct ioloop for client AsyncEvent objects

ajoubertza:
A number of `AsyncEvent` objects are created in the client's `__init__` method, resulting in them getting a reference to the current ioloop at the time of instantiation.  If a different ioloop is used when starting the client (which will happen in the "managed" case), this can cause problems.  E.g., if the original ioloop is closed before stopping the client, an exception occurs during
`yield self._stopped.until_set`.  This was only noticed after the changes in PR #206.

The fix is to recreate the `AsyncEvent` objects using the correct ioloop when starting the client.  Some of the other state variables that should have been cleared after stopping, for example the protocol received, are now also re-initialised each time.

Problems with `AsyncEvent` objects initialised in other higher level classes were not fixed, but should be considered for future work.

JIRA: [MT-834](https://skaafrica.atlassian.net/browse/MT-834)

### DIFF
--- a/katcp/client.py
+++ b/katcp/client.py
@@ -147,46 +147,63 @@ class DeviceClient(object):
         self._bindaddr = (host, port)
         self._tb_limit = tb_limit
         self._logger = logger
-        # Indicate that client's main coroutine is not running
-        self._stopped = AsyncEvent()
-        self._stopped.set()
-        # Indicate that client is running and ready connect
-        self._running = AsyncEvent()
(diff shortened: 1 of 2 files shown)


bmerry:
@ajoubertza I was happy to help out with hunting memory leaks, but I don't think it makes sense for me to be a general python-katcp reviewer (except where it affects aiokatcp), given that SDP doesn't use python-katcp and I'm not very familiar with the interface, the implementation or Tornado.

PR #211, opened 2019-09-03
Futurize katcp files

tockards:
Migrated using futurize stage 2 and omitted the `from builtins import *` and the `from builtins import unicode_literals`

- inspecting_client.py
- ioloop_manager.py
- kattypes.py
- object_proxies.py
- resource.py
- resource_client.py
- sensortree.py
- object_proxies.py


mmphego:
LGTM

PR #217, opened 2019-09-16
Fixed imports and licencing issues

amakhaba:
The work done in this PR fixes imports and licensing related issues

JIRA: [MT-737](https://skaafrica.atlassian.net/browse/MT-737)

### DIFF
--- a/bench/runner.py
+++ b/bench/runner.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright 2009 National Research Foundation (South African Radio Astronomy Observatory)
+# Copyright 2010 National Research Foundation (South African Radio Astronomy Observatory)
 # BSD license - see LICENSE for details
 
 from __future__ import absolute_import, division, print_function
--- a/bench/runner.py
+++ b/bench/runner.py
@@ -2,14 +2,17 @@
 # Copyright 2009 National Research Foundation (South African Radio Astronomy Observatory)
 # BSD license - see LICENSE for details
 
-from __future__ import print_function
-from __future__ import division
-from __future__ import absolute_import
+from __future__ import absolute_import, division, print_function
+
+import os
+import re
+import signal
+import sys
+from optparse import OptionParser
+
 from twisted.internet import reactor
 from twisted.internet.protocol import ProcessProtocol
-from optparse import OptionParser
 from twisted.python import log
-import re, os, sys, signal
 
 log.startLogging(open("log", "w"), setStdout=False)
 
--- a/doc/exts/docscrape.py
+++ b/doc/exts/docscrape.py
@@ -1,14 +1,12 @@
 """Extract reference documentation from the NumPy source tree.
 
 """
-from __future__ import print_function
-from __future__ import division
-from __future__ import absolute_import
+from __future__ import absolute_import, division, print_function
 
 import inspect
-import textwrap
-import re
 import pydoc
+import re
+import textwrap
 from StringIO import StringIO
 from warnings import warn
 
--- a/katcp/core.py
+++ b/katcp/core.py
@@ -6,30 +6,31 @@
 
 """Utilities for dealing with KAT device control language messages."""
 
-from __future__ import division, print_function, absolute_import
-
+from __future__ import absolute_import, division, print_function
+import future
 from future import standard_library
-standard_library.install_aliases()
 
+import logging
 import re
 import sys
 import time
 import warnings
-import logging
-
-import future
-import tornado
 
 from builtins import bytes, next, object, range, str
-from past.builtins import basestring
-from collections import namedtuple, defaultdict
-from functools import wraps, partial
-
+from collections import defaultdict, namedtuple
 from concurrent.futures import Future, TimeoutError
-from tornado import gen
-from tornado.gen import with_timeout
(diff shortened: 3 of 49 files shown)


bench/runner.py
```diff
@@ -2,14 +2,17 @@
 # Copyright 2009 National Research Foundation (South African Radio Astronomy Observatory)
```
ajoubertza: Please check all these new copyright headers that have been added - the year must match when the file was first created.  Looking at the [history](https://github.com/ska-sa/katcp-python/commits/2ddcc4589ed9a3060c8d6ce2910beebce44c8a68/bench/runner.py) of this file, it only appeared in 2010.

doc/exts/docscrape.py
```diff
@@ -1,14 +1,12 @@
...
-from __future__ import absolute_import
+from __future__ import absolute_import, division, print_function
 
 import inspect
-import textwrap
-import re
 import pydoc
+import re
+import textwrap
 from StringIO import StringIO
```
mmphego: CR would be nice here, to divide `imports` and `from`

katcp/core.py
```diff
@@ -1,35 +1,32 @@
...
-# Copyright 2009 National Research Foundation (South African Radio Astronomy Observatory)
+# Copyright 2010 National Research Foundation (South African Radio Astronomy Observatory)
 # BSD license - see LICENSE for details
 
 """Utilities for dealing with KAT device control language messages."""
 
-from __future__ import division, print_function, absolute_import
-
+from __future__ import absolute_import, division, print_function
+import future
```
mmphego: `future` is a 3rd party module, perhaps place it under 3rd party imports. Do not confuse `from future import standard_library` and `future`...

Read here: https://python-future.org/reference.html?highlight=standard_library#id13

katcp/object_proxies.py
```diff
@@ -1,20 +1,15 @@
-# Copyright 2014 National Research Foundation (South African Radio Astronomy Observatory)
```
mmphego: Any reason as to why you remove the license header?

katcp/server.py
```diff
@@ -6,54 +6,48 @@
 
 """Servers for the KAT device control language."""
 
-from __future__ import division, print_function, absolute_import
-
+from __future__ import absolute_import, division, print_function
+import future
```
mmphego: See above comment

katcp/test/test_sensortree.py
```diff
@@ -1,10 +1,9 @@
-# Copyright 2009 National Research Foundation (South African Radio Astronomy Observatory)
+# Copyright 2010 National Research Foundation (South African Radio Astronomy Observatory)
```
mmphego: Any reason as to why you decided to change the year of the license?

katcp/test/test_testutils.py
```diff
@@ -1,21 +1,19 @@
-# Copyright 2009 National Research Foundation (South African Radio Astronomy Observatory)
+# Copyright 2012 National Research Foundation (South African Radio Astronomy Observatory)
```
mmphego: Ditto

scratchpad/tornado/test_tornado_client.py
```diff
@@ -1,15 +1,14 @@
...
-from tornado.util import ObjectDict
+from __future__ import absolute_import, division, print_function
+
 from tornado.concurrent import Future as tornado_Future
-from katcp import Sensor
+from tornado.util import ObjectDict
 
-from katcp.testutils import DeviceTestServer
+from katcp import Sensor
 from katcp.client import *
```
mmphego: **Note:** This should be avoided at all times ie `from xxx import *`

scripts/__init__.py
```diff
@@ -6,11 +6,9 @@
    @author Robert Crida <robert.crida@ska.ac.za>
    @date 2008-10-10
    """
-from __future__ import print_function
-from __future__ import division
-from __future__ import absolute_import
-from __future__ import unicode_literals
-
+from __future__ import (absolute_import, division, print_function,
```
mmphego: I don't think these imports are necessary!, this file can be left blank

katcp/core.py
ludwigschwardt: I prefer the current version, simply because it puts all the imports for Python 2 support in one place. One Day ™️ we can just rip out the entire code block in one go when we go Python 3 only.

mmphego: I suppose it would be much easier to rip them off when that One Day ™️. 
Thanks for the input.

doc/exts/docscrape.py
bmerry: Note that PEP 8 also makes [recommendations](https://www.python.org/dev/peps/pep-0008/#imports) about order and grouping of imports (which flake8 doesn't check unfortunately).

katcp/object_proxies.py
amakhaba: The file was sourced from somewhere else, and has its own CP right header

katcp/test/test_sensortree.py
amakhaba: @ajoubertza suggested that it would be better to look at the history of the file, then the CP should reflect when the file was created.

PR #218, opened 2019-09-20
Update KATCP guidelines PDF to Rev 5.1

ajoubertza:
The specification has been updated.  Note that not all of the new features have been implemented yet.


doc/index.rst
```diff
@@ -8,8 +8,9 @@ It is developed as a part of the `Karoo Array Telescope <http://ska.ac.za/>`_
 (KAT) project and used at KAT for the monitoring and control of hardware
 devices.
 
-The protocol specification :download:`NRF-KAT7-6.0-IFCE-002-Rev5.pdf`
-is maintained as an internal memo. The latest version is `Rev5`.
+The protocol specification :download:`NRF-KAT7-6.0-IFCE-002-Rev5-1.pdf`
+is maintained as an internal memo. The latest version is `Rev5.1`.
+The specification source is hosted here:  https://github.com/ska-sa/katcp-guidelines
```
bmerry: Will sphinx automatically turn that URL into a link?

bmerry:
Can you keep the previous PDF around (even if it's not referenced) to avoid breaking links and for historical reference? aiokatcp currently points at it for example.

doc/index.rst
ajoubertza: Yes, plain text like this does become a hyperlink (at least on readthedocs).

PR #219, opened 2019-09-23
Added Sensor.string test, to check if sensor string type is of f…

mmphego:
Added additional tests.

### DIFF
--- a/katcp/core.py
+++ b/katcp/core.py
@@ -988,6 +988,11 @@ class Sensor(object):
         if initial_status is None:
             initial_status = Sensor.UNKNOWN
 
+        # Checking if sensor_type is a native string,
+        # if it is then assign it to a PY3 backport str object
+        if sensor_type == future.utils.native_str:
+            sensor_type = future.types.newstr
+
         sensor_type = self.SENSOR_SHORTCUTS.get(sensor_type, sensor_type)
 
         self._sensor_type = sensor_type
--- a/katcp/test/test_core.py
+++ b/katcp/test/test_core.py
@@ -17,6 +17,7 @@ import unittest
 
 from builtins import object, str
 
+import future
 import tornado
 
 import katcp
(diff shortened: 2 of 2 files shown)


katcp/core.py
```diff
@@ -988,6 +988,11 @@ def __init__(self, sensor_type, name, description=None, units='',
         if initial_status is None:
             initial_status = Sensor.UNKNOWN
 
+        # Checking if sensor_type is a native string,
+        # if it is then assign it to a PY3 backport str object
+        if sensor_type == future.utils.native_str:
```
mmphego: @ajoubertza did you do similar checks on `core.py`?

ajoubertza: No.  I didn't notice that part yet.  I see there could be an issue if someone tries to instantiate `Sensor` objects directly, rather than using the convenience constructors like `sensor.string()`, which use the `Sensor.STRING` constant for the type.

In my branch, `Sensor("blah", ...)` would be fine in PY2 and PY3 (native string).  Using `Sensor(u"blah", ...)` would be a problem in PY2, but OK in PY3.  I'm guessing none of our PY2 code is instantiating with unicode strings, so no problems at the moment.

I'm busy thinking about the `kattypes.Str`, and how it would work with `String` sensors and as request/reply arguments when unicode strings get into the mix...

mmphego: > No. I didn't notice that part yet. I see there could be an issue if someone tries to instantiate `Sensor` objects directly, rather than using the convenience constructors like `sensor.string()`, which use the `Sensor.STRING` constant for the type.

In an ideal environment would we need to care about the type of string it is? whether it's a `native_str` or `newstr`?

lanceWilliams: Not sure I follow the functioning of this check. The doc seemed to indicate that native_str just resolves to whatever the builtin is (which I assume means the version that is running this)? Shouldn't the goal be to cast to unicode internally regardless of what is coming in?

PR #226, opened 2019-11-04
Updated license header on files.

mmphego:
Renamed license header from *SKA South Africa* to *National Research Foundation (South African Radio Astronomy Observatory)*, which is consistent across most of the files.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -11,9 +11,8 @@ modification, are permitted provided that the following conditions are met:
     * Redistributions in binary form must reproduce the above copyright notice,
       this list of conditions and the following disclaimer in the documentation
       and/or other materials provided with the distribution.
-
-    * Neither the name of the National Research Foundation
-      (South African Radio Astronomy Observatory) project nor the names of its
+    * Neither the name of the National Research Foundation or
+      South African Radio Astronomy Observatory nor the names of its
       contributors may be used to endorse or promote products derived from this
       software without specific prior written permission.
 
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,5 @@
-Copyright (c) 2009, 2010 National Research Foundation (South African Radio Astronomy Observatory)
+Copyright (c) 2009, 2010 National Research Foundation
+(South African Radio Astronomy Observatory)
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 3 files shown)


LICENSE
```diff
@@ -11,7 +11,7 @@ modification, are permitted provided that the following conditions are met:
       this list of conditions and the following disclaimer in the documentation
       and/or other materials provided with the distribution.
 
-    * Neither the name of the SKA South Africa project nor the names of its
+    * Neither the name of the National Research Foundation (South African Radio Astronomy Observatory) project nor the names of its
```
amakhaba: Apart from jenkins complaining about the length of the line, Looks Ok

ajoubertza: Agree it would be better to split this line.  It isn't a Python file, but still nicer not to have very long lines.

doc/conf.py
```diff
@@ -47,7 +47,7 @@
 
 # General information about the project.
 project = u'KATCP'
-copyright = u'2009-2012, SKA South Africa'
+copyright = u'2009-2012, National Research Foundation (South African Radio Astronomy Observatory)'
```
ajoubertza: In the other files we don't have an upper bound on the the copyright.  Either extend the upper limit to 2019, or remove it (the latter is easier to maintain).

ludwigschwardt: Interesting point. Copyright only seems to care about the _first_ year of publication, so it does indeed sound like a great idea to remove those pesky upper bounds 😄

LICENSE
```diff
@@ -11,7 +12,8 @@ modification, are permitted provided that the following conditions are met:
       this list of conditions and the following disclaimer in the documentation
       and/or other materials provided with the distribution.
 
-    * Neither the name of the SKA South Africa project nor the names of its
+    * Neither the name of the National Research Foundation
+      (South African Radio Astronomy Observatory) project nor the names of its
```
ludwigschwardt: I would rather say something like:

`Neither the name of the National Research Foundation or South African Radio Astronomy Observatory nor the names...`

mmphego:
> Thanks for noticing these.

Thanks to @amakhaba for picking it up. (https://github.com/ska-sa/katmisc/pull/152#discussion_r342010374)

PR #227, opened 2019-12-23
Removed pip's index and trusted hosts

mmphego:
This PR does the following:

- Removes pip index and trusted hosts related information from `tox.ini` file.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,9 +3,6 @@ envlist = py{27, 36},docs
 
 [testenv]
 passenv = test_flags
-install_command = pip install \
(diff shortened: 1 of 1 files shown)


tockards:
@mmphego 
I take it that we are removing it from tox because it is now in the container?
in its .pip or w/e configuration file it suppose to be in.

mmphego:
> @mmphego
> I take it that we are removing it from tox because it is now in the container?
> in its .pip or w/e configuration file it suppose to be in.

Yeah, plus also for end-users who would pro'ly want to run tests and not have access to our PyPI...

PR #232, opened 2020-04-15
Pinned mock and coverage dependencies.

mmphego:
This PR does the following:
- Pins testing dependencies due to the latest version of Mock causes python3.xx tests to fail: http://ci.camlab.kat.ac.za/job/katcp-multibranch/job/master/31/console


tockards:
@ajoubertza Yes, I had to remove it as flake8 was complaining and brought it back under the method docstring with information of its context.

PR #233, opened 2020-04-23
ensure tested requests are byte strings

tockards:
This PR  increases the test utils stability by ensuring that all  [`test_help`](https://github.com/ska-sa/katcp-python/blob/2b9da670dc05493a3f2f5b1832f6d92172e6b976/katcp/testutils.py#L668) requests compared in the tests are bytes. 
JIRA: [P2M-17](https://skaafrica.atlassian.net/browse/P2M-17)

### DIFF
--- a/katcp/test/test_server.py
+++ b/katcp/test/test_server.py
@@ -25,7 +25,7 @@ import _thread
 import mock
 import tornado.testing
 
-from future.utils import PY2
+from future.utils import PY2, PY3
 from tornado import gen
 
 import katcp
@@ -1080,7 +1080,7 @@ class TestDeviceServerClientIntegrated(unittest.TestCase, TestUtilMixin):
 
     def test_test_help_failure(self):
         """Basic Test to exercise test_help function
-        failure by supplying byte strings.
+        failure by supplying byte strings in Python 3.
         """
         request_names = [
             b"cancel-slow-command",
@@ -1091,8 +1091,9 @@ class TestDeviceServerClientIntegrated(unittest.TestCase, TestUtilMixin):
             b"raise-fail",
             b"slow-command",
         ]
-        with self.assertRaises(ValueError):
-            self.client.test_help(request_names)
+        if PY3:
+            with self.assertRaises(ValueError):
+                self.client.test_help(request_names)
 
 
 class TestHandlerFiltering(unittest.TestCase):
--- a/katcp/test/test_server.py
+++ b/katcp/test/test_server.py
@@ -1067,14 +1067,33 @@ class TestDeviceServerClientIntegrated(unittest.TestCase, TestUtilMixin):
     def test_test_help(self):
         """Basic Test to exercise test_help function
         """
-        request_names = ['cancel-slow-command',
-                         'decorated',
-                         'decorated-return-exception',
-                         'new-command',
-                         'raise-exception',
-                         'raise-fail',
-                         'slow-command']
+        request_names = [
+            "cancel-slow-command",
+            "decorated",
+            "decorated-return-exception",
+            "new-command",
+            "raise-exception",
+            "raise-fail",
+            "slow-command",
+        ]
         self.client.test_help(request_names)
+
+    def test_test_help_failure(self):
+        """Basic Test to exercise test_help function
+        failure by supplying byte strings.
+        """
+        request_names = [
+            b"cancel-slow-command",
+            b"decorated",
+            b"decorated-return-exception",
+            b"new-command",
+            b"raise-exception",
+            b"raise-fail",
+            b"slow-command",
+        ]
+        with self.assertRaises(ValueError):
+            self.client.test_help(request_names)
+
 
 class TestHandlerFiltering(unittest.TestCase):
     class DeviceWithEverything(katcp.DeviceServer):
--- a/katcp/test/test_server.py
+++ b/katcp/test/test_server.py
@@ -1094,7 +1094,9 @@ class TestDeviceServerClientIntegrated(unittest.TestCase, TestUtilMixin):
         if PY3:
             with self.assertRaises(ValueError):
                 self.client.test_help(request_names)
-
+        # Please note that this does not fail in python2
+        if PY2:
+            self.client.test_help(request_names)
 
 class TestHandlerFiltering(unittest.TestCase):
     class DeviceWithEverything(katcp.DeviceServer):
--- a/katcp/test/test_server.py
+++ b/katcp/test/test_server.py
@@ -1064,6 +1064,17 @@ class TestDeviceServerClientIntegrated(unittest.TestCase, TestUtilMixin):
             # Use the tornado future to get a usable traceback
             tornado_future.result()
 
+    def test_test_help(self):
+        """Basic Test to exercise test_help function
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -57,6 +57,7 @@
     install_requires=[
         "ply",
         "future",
+        "six",
```
ajoubertza, line 60: I tried to avoid using `six` in the earlier work to make katcp Python 2 and 3 compatible. I didn't want to use two different libraries (future and six) for the compatibility.  Was that a reasonable thought?

I'm sure you spotted `future.utils.tobytes`, which is almost what we want, except it uses `latin-1` instead of `utf-8` encoding.

What about adding something like this to the `katcp.compat` module?
```python
def ensure_byte_str(value):
    """Coerce unicode string to bytes (UTF-8 encoding), no change if input is bytes."""
    if is_bytes(value):
        return value
    elif is_text(value):
        return value.encode('utf-8')
    else:
        raise TypeError(
                "Invalid type for string conversion: {}".format(type(value)))
```

katcp/testutils.py
```diff
@@ -699,12 +701,12 @@ def test_help(self, expected_requests, full_descriptions=False,
 
         if descriptions:
             if not full_descriptions:
-                got_requests = [(name, desc.split('\n')[0])
+                got_requests = [(ensure_binary(name), ensure_binary(desc).split(b'\n')[0])
```
ajoubertza, line 704: We should not do this for the `got_requests`.  Those are returned by other katcp code (the message `arguments`), and must already be byte strings.  It is only the user input, i.e., `expected_requests`, that we are unsure about.

katcp/testutils.py
```diff
@@ -699,12 +701,12 @@ def test_help(self, expected_requests, full_descriptions=False,
...
                                 for (name, desc) in got_requests]
-                expected_requests = [(name, desc.split('\n')[0])
+                expected_requests = [(ensure_binary(name), ensure_binary(desc).split(b'\n')[0])
                                      for (name, desc) in expected_requests]
         else:
-            got_requests = [name for (name, desc) in got_requests]
+            got_requests = [ensure_binary(name) for (name, desc) in got_requests]
```
ajoubertza, line 709: Ditto - shouldn't need this.

setup.py
tockards: cool, I will future. I used six as it was already installed on the docker container.

katcp/testutils.py
ajoubertza: To be more clear:  we do need the `b` prefix for the split, but not the `ensure_binary` calls.

setup.py
tockards: I was thinking of using `https://python-future.org/reference.html?highlight=to_bytes#future.utils.native_str_to_bytes` but let me rather add the above code to the `compat.py`

katcp/testutils.py
```diff
@@ -18,6 +18,8 @@
 import time
 
 from builtins import next, object, zip
+from six import ensure_binary
```
ajoubertza, line 21: Please remove.

setup.py
ajoubertza: Oh, yes.  `native_str_to_bytes` could also work - I didn't see that one.  As long as we are sure that the caller is using native strings.  That seems pretty reasonable, since it is the most convenient form.

ludwigschwardt: Was gonna suggest the same :-)

katcp/testutils.py
```diff
@@ -699,10 +701,13 @@ def test_help(self, expected_requests, full_descriptions=False,
...
                                 for (name, desc) in got_requests]
-                expected_requests = [(name, desc.split('\n')[0])
-                                     for (name, desc) in expected_requests]
+                expected_requests = [
+                    (native_str_to_bytes(name),
+                     native_str_to_bytes(desc).split(b"\n")[0])
+                    for (name, desc) in expected_requests
+                ]
         else:
             got_requests = [name for (name, desc) in got_requests]
```
ajoubertza, line 712: I think we also need to convert the expected requests if we're not doing descriptions.
```suggestion
            got_requests = [name for (name, desc) in got_requests]
            expected_requests = [native_str_to_bytes(name) for name in expected_requests]

```
Suggestion untested, so please try it out.

ajoubertza:
@tockards Can you also add a line to [CHANGELOG.md](https://github.com/ska-sa/katcp-python/blob/master/CHANGELOG.md), to make it easier when we get to the next release.  Something like this:
```
(Unreleased)
- Something about testutils...

28 October 2019 (0.7.0)
- Added Python 3 compatibility (PR #215)
```

katcp/testutils.py
```diff
@@ -699,19 +701,28 @@ def test_help(self, expected_requests, full_descriptions=False,
...
         got_set = set(got_requests)
         expected_set = set(expected_requests)
 
         if exclude_defaults:
             default_request_names = set(DeviceServer._request_handlers.keys())
+            bytes_str_default_request_names = set()
+            for name in default_request_names:
+                bytes_str_default_request_names.add(native_str_to_bytes(name))
 
+            default_request_names = bytes_str_default_request_names
```
ajoubertza, line 725: How about doing all this using a set comprehension instead?
```python
default_request_names = {native_str_to_bytes(name) for name in DeviceServer._request_handlers.keys()}
```

Or using map:
```python
default_request_names = set(map(native_str_to_bytes, DeviceServer._request_handlers.keys())
```

(Untested)

tockards: I wanted to be a bit explicit for future persons debugging this as it is test code and not production code but I will add the set comprehension. 
I also noticed that the tests have been failing since this change so I have to look into it.

ajoubertza: Thanks.  I think that Python 3.8 failure is a flaky test.  I ran it again, and it passed.  Now coverage is indicating a problem?

katcp/test/test_server.py
```diff
@@ -1064,6 +1064,17 @@ def do_async_request_test(threadsafe_future, tornado_future):
...
+        """Basic Test to exercise test_help function
+        """
+        request_names = ['cancel-slow-command',
+                         'decorated',
+                         'decorated-return-exception',
+                         'new-command',
+                         'raise-exception',
+                         'raise-fail',
+                         'slow-command']
+        self.client.test_help(request_names)
```
mmphego, line 1077: Please run black, it should fix the formatting.

ajoubertza, line 1077: Test for the positive case is fine.  Please add a test for failure as well.

ajoubertza:
@tockards Looks like the drop in coverage is causing the CI run to fail.  It is still higher than the last master run, so this doesn't block the PR for me.

```
...
[Pipeline] cobertura
[Cobertura] Publishing Cobertura coverage report...
[Cobertura] Publishing Cobertura coverage results...
[Cobertura] Cobertura coverage report found.
[Cobertura] Code coverage enforcement failed for the following metrics:
[Cobertura]     Lines's stability is 89.31 and set mininum stability is 91.23.
```

tockards:
@ajoubertza Thanks. It might be that i made the test for failure not work in python2.

PR #234, opened 2020-05-05
Fixes to KATCP Types codecs w.r.t Python 2 and Python 3 

tockards:
Opening the PR for comments. This PR was created due address issues that were encountered with katcp during the Python 2 to Python 3 migration of [mkat-tango](https://github.com/ska-sa/mkat-tango)

This PR addresses issues with the encoding and decoding of sensor values w.r.t KATCP Types and Python 2 and Python 3 compatibility. 

- More tests were added for sensors that weren't tested
- This adds some conflicting Changes to functionality of  `Str` decode in Python 2 and Python 3

TODO
- [ ] Update CHANGES.md once PR is OKed 

PLEASE NOTE
 - There are currently issues with the CI job of this PR that is related to the code coverage.

JIRA: [P2M-17](https://skaafrica.atlassian.net/browse/P2M-17)

### DIFF
--- a/katcp/kattypes.py
+++ b/katcp/kattypes.py
@@ -235,10 +235,10 @@ class Bool(KatcpType):
     encode = lambda self, value, major: value and b"1" or b"0"
 
     def decode(self, value, major):
-        if value not in (b"0", b"1", "0", "1"):
+        if value not in (b"0", b"1"):
             raise ValueError("Boolean value must be b'0' or b'1' but is '%s'."
                              % (value,))
-        return value == b"1" or value == "1"
+        return value == b"1"
 
 
 class Str(KatcpType):
--- a/katcp/kattypes.py
+++ b/katcp/kattypes.py
@@ -394,6 +394,9 @@ class Address(KatcpType):
         return b"%s:%d" % (host, port) if port is not None else host
 
     def decode(self, value, major):
+        if is_text(value):
+            if not future.utils.PY2:
+                value = future.utils.native_str_to_bytes(value)
         if value.startswith(b"["):
             match = self.IPV6_RE.match(value)
         else:
--- a/katcp/kattypes.py
+++ b/katcp/kattypes.py
@@ -286,7 +286,12 @@ class Str(KatcpType):
                 return str(value).encode('utf-8')
 
         def decode(self, value, major):
-            return value.decode('utf-8')
+            if is_bytes(value):
+                return value.decode('utf-8')
+            elif is_text(value):
+                return value
+            else:
+                return str(value)
 
 
 class Discrete(Str):
--- a/katcp/kattypes.py
+++ b/katcp/kattypes.py
@@ -342,8 +342,8 @@ class Lru(KatcpType):
 
     ## @brief Mapping from LRU value constant to LRU value name.
     LRU_VALUES = {
-        LRU_NOMINAL: "nominal",
-        LRU_ERROR: "error",
+        LRU_NOMINAL: b"nominal",
+        LRU_ERROR: b"error",
     }
 
     # LRU_VALUES not found by pylint
@@ -358,6 +358,9 @@ class Lru(KatcpType):
         return Lru.LRU_VALUES[value]
 
     def decode(self, value, major):
+        if is_text(value):
+            if future.utils.PY3:
+                value = future.utils.native_str_to_bytes(value)
         if value not in Lru.LRU_CONSTANTS:
             raise ValueError("Lru value must be 'nominal' or 'error'.")
         return Lru.LRU_CONSTANTS[value]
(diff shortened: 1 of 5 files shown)


katcp/kattypes.py
```diff
@@ -235,10 +235,10 @@ class Bool(KatcpType):
     encode = lambda self, value, major: value and b"1" or b"0"
 
     def decode(self, value, major):
-        if value not in (b"0", b"1"):
+        if value not in (b"0", b"1", "0", "1"):
             raise ValueError("Boolean value must be b'0' or b'1' but is '%s'."
                              % (value,))
-        return value == b"1"
+        return value == b"1" or value == "1"
```
bmerry, line 241: It's been a long time since I used this package so maybe I'm missing some context about how kattypes is used, but this seems a bit odd. I can understand decoded strings being either bytes or str because the katcp protocol has only a single type for text and raw data. But encoding should always produce bytes (because that's what sockets work on) and hence decoding should always expect bytes, not well-maybe-its-bytes-or-maybe-its-not. It smells like the calling code isn't  keeping a clean separation between encoded data on the wire and decoded text.

ajoubertza: @bmerry Your memory (and nose) is spot on.

The specific method that triggered this change was `katcp.Sensor.parse_params`, which uses the sensor's underlying type to decode the parameters.  In this context, the word `formatted` in the variable name indicates that it should be "formatted" like it would be serialised on the wire (i.e., byte strings).
https://github.com/ska-sa/katcp-python/blob/df7f3a425c6387b9b3c125d5db7c37fdf748a472/katcp/core.py#L1568

When we use it from the inspecting client, we pass a `sensor_info` object that will have native strings instead of byte strings.  The change to native strings was part of the previous round of migration to Python 3.
https://github.com/ska-sa/katcp-python/blob/df7f3a425c6387b9b3c125d5db7c37fdf748a472/katcp/inspecting_client.py#L872-L874

Yes, we could fix the usages in inspecting client (and fake inspecting client) instead.  I'm not sure if there are other bad usages of `decode`, but I imagine they will show up easily enough.

tockards: @bmerry and @ajoubertza 
I addressed the issues with parsing parameters for boolean sensors where `parse_params` is called.
it is a bit hacky so any input if and when you are available would be appreciated.

katcp/test/test_fake_clients.py
```diff
@@ -35,6 +35,12 @@ def test_sensors(self):
         sensor_info = {
             'an-int': ('An integer sensor', 'things', 'integer', 0, 10),
             'a-string' : ('A string sensor', '', 'string'),
+            'a-discrete': ('A discrete sensor', '', 'discrete', 'a', 'b', 'c'),
+            'a-timestamp': ('A timestamp sensor', '', 'timestamp', 1556928000),
+            'a-float': ('A float sensor', '', 'float', 1234.42),
+            'a-boolean': ('A boolean sensor', '', 'boolean', '0'),
+            'an-address': ('An address sensor', '', 'address', '127.0.0.1:8080'),
+            'a-lru': ('A LRU sensor', '', 'lru', 'nominal'),
```
ajoubertza, line 43: Some of these test values are not valid.  The params field isn't used for all types of sensors, so some of the tuples should only have 3 items.  For floats, params are optional, but we need at least 2: min and max.  From [KATCP guidelines](https://github.com/ska-sa/katcp-python/blob/master/doc/NRF-KAT7-6.0-IFCE-002-Rev5-1.pdf):

image

```suggestion
            'a-discrete': ('A discrete sensor', '', 'discrete', 'a', 'b', 'c'),
            'a-timestamp': ('A timestamp sensor', '', 'timestamp'),
            'a-float': ('A float sensor', '', 'float', 1234.42, 5678.99),
            'a-boolean': ('A boolean sensor', '', 'boolean'),
            'an-address': ('An address sensor', '', 'address'),
            'a-lru': ('A LRU sensor', '', 'lru'),
```

katcp/test/test_fake_clients.py
```diff
@@ -51,6 +57,66 @@ def test_sensors(self):
         self.assert_sensor_equal_description(a_string, dict(
```
ajoubertza, line 57: Since the new assertions use literal dicts (to make the linter happy), suggest the old two test do the same for consistency.

katcp/test/test_fake_clients.py
```diff
@@ -51,6 +57,66 @@ def test_sensors(self):
...
+            },
+        )
+        a_timestamp = yield self.fake_inspecting_client.future_get_sensor("a-timestamp")
+        self.assert_sensor_equal_description(
+            a_timestamp,
+            {
+                "name": "a-timestamp",
+                "type": Sensor.TIMESTAMP,
+                "description": "A timestamp sensor",
+                "params": [1556928000],
```
ajoubertza, line 77: I think the params will come out as `None`.

tockards:
Closing this PR as the plan is to not change kattypes.

PR #237, opened 2020-06-15
Support for handling generator expressions in Discrete type

mmphego:
This PR does the following:

- [x] Added support for handling generator expressions in `Discrete` type. 

This fixes: https://github.com/ska-sa/katcore/blob/4db3ec97d140ba4917be68d2335660e7f06baa1d/katcore/sim_base.py#L161

JIRA: [MT-1543](https://skaafrica.atlassian.net/browse/MT-1543)

### DIFF
--- a/katcp/kattypes.py
+++ b/katcp/kattypes.py
@@ -306,7 +306,7 @@ class Discrete(Str):
         super(Discrete, self).__init__(**kwargs)
         self._case_insensitive = case_insensitive
         if not isinstance(values, list):
-            raise ValueError("%r values parsed should be a list, instead of %r" % (
+            raise ValueError("%r values passed should be a list, instead of %r" % (
                 values, type(values))
             )
         self._values = list(values)  # just to preserve ordering
--- a/katcp/kattypes.py
+++ b/katcp/kattypes.py
@@ -305,8 +305,12 @@ class Discrete(Str):
     def __init__(self, values, case_insensitive=False, **kwargs):
         super(Discrete, self).__init__(**kwargs)
         self._case_insensitive = case_insensitive
+        if not isinstance(values, list):
+            raise ValueError("%r values parsed should be a list, instead of %r" % (
+                values, type(values))
+            )
         self._values = list(values)  # just to preserve ordering
-        self._valid_values = set(values)
+        self._valid_values = set(self._values)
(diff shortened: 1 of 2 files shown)


katcp/kattypes.py
```diff
@@ -305,8 +305,12 @@ class Discrete(Str):
     def __init__(self, values, case_insensitive=False, **kwargs):
         super(Discrete, self).__init__(**kwargs)
         self._case_insensitive = case_insensitive
+        if not isinstance(values, list):
+            raise ValueError("%r values parsed should be a list, instead of %r" % (
```
bmerry, line 309:
```suggestion
            raise ValueError("%r values passed should be a list, instead of %r" % (
```

bmerry:
This doesn't seem like a very Pythonic change - it looks like you've fixed the code so that it will work with any iterable, and hence forcing the caller to make a list just makes the function less friendly.

katcp/kattypes.py
```diff
@@ -305,8 +305,12 @@ class Discrete(Str):
     def __init__(self, values, case_insensitive=False, **kwargs):
         super(Discrete, self).__init__(**kwargs)
         self._case_insensitive = case_insensitive
+        if not isinstance(values, list):
```
ajoubertza, line 308: Do we really need to be this strict?  A tuple or similar sequence would also be fine any.  The casting to `list(values)` would work with all of those.  Even with a generator, after the change on line 313

I.e., follow the duck-typing approach that Python encourages (at least before mypy... 🙂 ) .

mmphego: Mmmhh yeah a bit too strict from my end, but as per the docstring https://github.com/ska-sa/katcp-python/blob/653d69e7d23c977bbb8d263380ab0d1aec5134a0/katcp/kattypes.py#L297 values should a list.

Will update the code. Thanks.

bmerry: Docstrings can be fixed :-) 

> I.e., follow the duck-typing approach that Python encourages (at least before mypy... 

Mypy supports duck-typing by means of "protocols" e.g. one could annotate this argument as `Iterable[str]`.

ajoubertza: Please update the PR title and description as well.  The main reason for doing this was to handle generators.

PR #239, opened 2020-07-23
Bytes  status in get_sensor to str

tockards:
This PR ensures that status in the `get_sensor` is str and not bytes. 
This was picked up in the migration of katcore from python 2 to python 3.

related JIRA : https://skaafrica.atlassian.net/browse/P2M-31

### DIFF
--- a/katcp/testutils.py
+++ b/katcp/testutils.py
@@ -28,7 +28,7 @@ import tornado.locks
 import tornado.testing
 from _thread import get_ident
 from tornado.concurrent import Future as tornado_Future
-from future.utils import native_str_to_bytes, bytes_to_native_str  # noqa: E402
+from future.utils import native_str_to_bytes, bytes_to_native_str
 
 from katcp import client
 
--- a/katcp/testutils.py
+++ b/katcp/testutils.py
@@ -28,7 +28,7 @@ import tornado.locks
 import tornado.testing
 from _thread import get_ident
 from tornado.concurrent import Future as tornado_Future
-from future.utils import native_str_to_bytes
+from future.utils import native_str_to_bytes, bytes_to_native_str  # noqa: E402
 
 from katcp import client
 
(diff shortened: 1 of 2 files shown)


katcp/testutils.py
```diff
@@ -28,7 +28,7 @@
 import tornado.testing
 from _thread import get_ident
 from tornado.concurrent import Future as tornado_Future
-from future.utils import native_str_to_bytes
+from future.utils import native_str_to_bytes, bytes_to_native_str  # noqa: E402
```
ajoubertza, line 31: E402 is "Module level import not at top of file", so I wonder why this was necessary.

tockards: Any time we add an import to the Python 2 Python 3 files that have `standard_library.install_aliases() ` above the other imports we run into this issue and I have to silence it by adding that `#noqa: E402` or disabling the linting. 
See the then jenkins job below that is failing. 
http://ci.camlab.kat.ac.za/job/katcp-multibranch/job/user%252Ftockards%252FP2M-31%252Fstatus-to-str/2/execution/node/22/log/

The other alternative is to ignore this within the `setup.cfg`

ajoubertza: Oh dear.  I wonder if it is because we are only trying to lint the diff, and the diff excludes the `standard_library.install_aliases()  # noqa: E402`?
Anyway, I think it would be better to disable this one globally, rather than having almost random `#noqa: E402` comments sprinkled throughout the codebase.

tockards:
It is currently failing due to documentation step.

ajoubertza:
> It is currently failing due to documentation step.

Looks like coverage failure.
```
[Cobertura] Code coverage enforcement failed for the following metrics:
[Cobertura]     Lines's stability is 89.79 and set mininum stability is 94.16.
```

tockards:
closing this PR as it will be addressed in another PR

PR #240, opened 2020-07-28
Fix get_sensor in testutils

tockards:
This PR addresses issues in `get_sensor` and adds a simple test that exercises the changes made to `get_sensor`

`status`  is returned from message arguments as a byte but then compares to a status that is a str and this changes fixes that.


tockards:
I am not sure why this happens `[Cobertura]     Lines's stability is 91.17 and set mininum stability is 91.65.` It might be that the Python 2 and Python 3 tests different parts of the code and ignores others. I have also noticed this 
https://github.com/jenkinsci/cobertura-plugin/issues/89
and its reference in the PR in katlogger https://github.com/ska-sa/katlogger/commit/26442eb011b5ddd66af4b721df6a173babe67fb7
Where we comment the katlogger coverage. 
What do I do to move forward?

tockards:
@ajoubertza I am not entirely sure why the cobertura line stability is persisting even tho the first test after this commit passed.

ajoubertza:
@tockards It seems like the Cobertura plugin is giving false positives.  I suggest disabling the minimum that allows us to continue.

setup.cfg
```diff
@@ -16,7 +16,7 @@ count = True
 ; D401 First line should be in imperative mood; try rephrasing
 ; W503: line break before binary operator (This is incompartible with Black,
 ;       See: https://black.readthedocs.io/en/stable/the_black_code_style.html#line-breaks-binary-operators)
-ignore = D401,W503
+ignore = D401,W503,E402
```
mmphego, line 19: Some info regarding the error we are ignoring.
```suggestion
; E402: Import at top, too many circular import fixes.
ignore = D401,W503,E402
```

Jenkinsfile
```diff
@@ -85,10 +85,12 @@ pipeline {
                         autoUpdateHealth: true,
                         autoUpdateStability: true,
                         zoomCoverageChart: true,
+                        //Ideally test coverage should be > 80%
                         lineCoverageTargets: '80, 80, 80',
                         conditionalCoverageTargets: '80, 80, 80',
                         classCoverageTargets: '80, 80, 80',
                         fileCoverageTargets: '80, 80, 80',
+                        methodCoverageTargets: '80, 0, 0',
```
mmphego, line 93: :+1:

tockards:
I have added `failBuildIfCoverageDecreasedInChangeRequest` and set it to `false` see https://www.jenkins.io/doc/pipeline/steps/code-coverage-api/

PR #241, opened 2020-08-06
only cast non bytes to str in params from assert_request_succeeds

tockards:
@ajoubertza I am currently migrating the katcore tests and I ran into something that appears to be another bug in katcp-python testutils for python2 and python3 .

It occurs in the testutils `assert_request_succeeds` method I added a line that exercises it for the case of passing an argument through that is a byte and similar to the tests found in katcore see https://github.com/ska-sa/katcore/blob/3b8f8f42192a47ed7e52f2f504787876ed6ed0f2/katcore/test/test_swcomp_base.py#L143

The PR is not a fix but a work around and aid to a potential fix as this bytes in the params might have to be caught earlier. 

Could you provide your input on this?

### DIFF
--- a/katcp/testutils.py
+++ b/katcp/testutils.py
@@ -8,7 +8,11 @@
 
 from __future__ import absolute_import, division, print_function
 from future import standard_library
+
 standard_library.install_aliases()  # noqa: E402
+
+from .compat import is_bytes
+
 
 import functools
 import logging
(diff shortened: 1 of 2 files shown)


katcp/testutils.py
```diff
@@ -792,7 +796,12 @@ def assert_request_succeeds(self, requestname, *params, **kwargs):
         informs_args_equal = kwargs.get("informs_args_equal")
 
         if args_echo:
-            args_equal = [str(p) for p in params]
```
tockards, line 795: This could also be fixed at line 785 see https://github.com/ska-sa/katcp-python/pull/241/files#diff-98b8b54ad65a0b5eb81c30e4ff66fa9aL785

tockards:
I am closing this PR in favour of following the approach in #242

PR #242, opened 2020-08-12
check if arg is bytes and then cast to str.

tockards:
This is an alternate approach to addressing the issue where byte strings are present in the arguments and causes issues down the line  where we do comparisons in the `assert_request_succeeds` method.

The other approach is https://github.com/ska-sa/katcp-python/pull/241 

related JIRA: https://skaafrica.atlassian.net/browse/P2M-34

### DIFF
--- a/katcp/test/test_server.py
+++ b/katcp/test/test_server.py
@@ -1014,7 +1014,7 @@ class TestDeviceServerClientIntegrated(unittest.TestCase, TestUtilMixin):
         self.server.ioloop.add_callback(self.server.clear_strategies, client_conn)
         self.server.sync_with_ioloop()
         self.client.assert_request_succeeds("sensor-sampling", "an.int",
-                                            args_equal=[b"an.int", b"none"])
+                                            args_equal=["an.int", "none"])
 
         # Check that we did not accidentally clobber the strategy datastructure
         # in the proccess
--- a/katcp/testutils.py
+++ b/katcp/testutils.py
@@ -38,6 +38,7 @@ from .kattypes import (Bool, Discrete, Float, Int, Str, concurrent_reply,
                        request, request_timeout_hint, return_reply)
 from .object_proxies import ObjectWrapper
 from .server import ClientConnection, DeviceServer, FailReply
+from .compat import is_bytes
 
 logger = logging.getLogger(__name__)
 
@@ -782,8 +783,11 @@ class BlockingTestClient(client.BlockingClient):
                                         "(with no error message)")))
         self.test.assertTrue(reply.reply_ok(), msg)
 
-        args = reply.arguments[1:]
-
+        args = []
+        for arg in reply.arguments[1:]:
+            if is_bytes(arg):
+                arg = bytes_to_native_str(arg)
+            args.append(arg)
         args_echo = kwargs.get("args_echo", False)
         args_equal = kwargs.get("args_equal")
         args_in = kwargs.get("args_in")


ajoubertza:
Discussed with @tockards - plan is to make `assert_request_succeeds` more lenient, to simplify the usage across Python versions.  It should accept byte-strings or native strings for the input parameters and the args to compare to. Both reply and inform arguments needs similar handling.  Coerce everything to native strings before comparing.

katcp/test/test_server.py
```diff
@@ -1014,13 +1014,59 @@ def test_sampling(self):
...
+        # the test client
+        client_conn = list(self.server._client_conns)[0]
+        self.server.ioloop.add_callback(self.server.clear_strategies, client_conn)
+        self.server.sync_with_ioloop()
+        self.client.assert_request_succeeds("sensor-sampling", "an.int",
+                                            args_equal=["an.int", "none"])
+        self.client.assert_request_succeeds("sensor-sampling", b"an.int",
+                                            args_equal=["an.int", "none"])
+        levels = ['off', b'fatal', 'error', b'warn', 'info', 'debug', 'trace', 'all']
+        self.client.assert_request_succeeds("log-level", args_in=[[level] for level in levels])
```
ajoubertza, line 1065: There are a few more conditions we should test:
- `args_equal` with byte strings
- `args_echo=True`
- `informs_args_equal` with byte strings and native strings
- that an exception is raised when `assert_request_succeeds` is used incorrectly - e.g., using it to call `?raise-fail`.

katcp/test/test_server.py
```diff
@@ -1014,13 +1014,59 @@ def test_sampling(self):
...
-                                            args_equal=[b"an.int", b"none"])
-
+                                            args_equal=["an.int", "none"])
         # Check that we did not accidentally clobber the strategy datastructure
         # in the proccess
         self.client.assert_request_succeeds(
             "sensor-sampling", "an.int", "period", 0.125)
 
+    def test_assert_request_succeeds(self):
+        """Test exercises assert_request_succeeds it is a copy of test_sampling."""
```
ajoubertza, line 1024: This tests seems overly complicated for testing `assert_request_succeeds`.  Could we simplify so that it is easier for the reader to see the required behaviour of `assert_request_succeeds`?
E.g., just using a standard request like `?sensor-value`, and removing all the sampling stuff.

tockards: Agreed. This was just the quickest for me as I was adding the test cases to the `test_sampling`. 
It seems that we have a similar set of issues with `assert_request_fails` see https://github.com/ska-sa/katcp-python/blob/0133e9ee0f5ba0cf7d5731105ad83592eacb178a/katcp/testutils.py#L829 
Should I follow the approach of this PR for `assert_request_fails`

ajoubertza: It could be added to this PR, but you can decide.

katcp/test/test_server.py
```diff
@@ -1014,13 +1014,26 @@ def test_sampling(self):
...
+        self.client.assert_request_succeeds("sensor-sampling", "an.int",
+                                            args_equal=["an.int", "none"])
+        self.client.assert_request_succeeds("sensor-sampling", b"an.int",
+                                            args_equal=["an.int", "none"])
+        self.client.assert_request_succeeds("sensor-sampling", b"an.int",
+                                            args_equal=[b"an.int", b"none"])
+        self.client.assert_request_succeeds("log-level", "debug", args_equal=["debug"],
+                                            args_echo=True)
+        self.client.assert_request_succeeds("log-level", "debug", args_equal=[b"debug"],
+                                            args_echo=True)
```
ajoubertza, line 1034: `args_echo` takes precedence over `args_equal` - which one do we want to test on these last two asserts?

katcp/test/test_server.py
```diff
@@ -1014,13 +1014,26 @@ def test_sampling(self):
...
+        levels = ['off', b'fatal', 'error', b'warn', 'info', 'debug', 'trace', 'all']
+        self.client.assert_request_succeeds("log-level", args_in=[[level] for level in levels])
```
ajoubertza, line 1037: Will tests still be added for these?
- `informs_args_equal` with byte strings and native strings
- testing that an exception is raised when `assert_request_succeeds` is used incorrectly - e.g., using it to call `?raise-fail`.

tockards: Yes this will be added.

tockards: @ajoubertza Should I add an exception that is raised when `assert_request_succeeds` is used incorrectly and add a test for that exception?

ajoubertza: Not sure if I understand the question.  Are you asking if we should add a new type of exception for this case?  Probably not.  An `AssertionError` seems like a reasonable choice for a test method called `assert_something`.  What does it raise?

katcp/test/test_server.py
```diff
@@ -1014,13 +1014,39 @@ def test_sampling(self):
...
+        self.client.assert_request_fails("sensor-sampling", "an.nonexistentsensor")
+        self.client.assert_request_fails(
+            "sensor-sampling",
+            "an.nonexistentsensor",
+            error_equals="Unknown sensor name: an.nonexistentsensor.",
+        )
+        self.client.assert_request_fails(
+            "sensor-sampling", "an.nonexistentsensor", status_equals="fail"
+        )
+
```
ajoubertza, line 1033: Similar to the `assert_request_suceeds` method, it would be good to test that an exception is raised in the negative case.  I.e., if a request that succeeds is sent via `assert_request_fails `.

ajoubertza (commented):
Just the negative tests still to be added.

Please update the PR title - is doesn't seem accurate anymore.

katcp/test/test_server.py
tockards: Ok, I was asking if I should throw exceptions in the `assert_request_suceeds` as there is it throws no exceptions. It raises nothing at the moment other than the exceptions you will encounter  in `assertEqual`

katcp/test/test_server.py
ajoubertza: > Ok, I was asking if I should throw exceptions in the assert_request_suceeds as there is it throws no exceptions. It raises nothing at the moment other than the exceptions you will encounter in assertEqual

No new exception to be added - the exception raised when `assertEqual` fails is perfect.  Or if we used a non-existent request, it would fail on this [`assertTrue`](https://github.com/ska-sa/katcp-python/blob/6ba34f3832ab8386546b4cdda5bff8d33d6ddbbc/katcp/testutils.py#L784).  I see you have added tests like that, which is what I was suggesting.

ajoubertza (approved):
LGTM

ajoubertza:
@tockards  Please update the PR title - is doesn't seem accurate anymore.

mmphego (approved):
Looks good but Jenkins not happy.

PR #243, opened 2020-08-14
use `bytes_to_native_str` inside `get_sensor` when sensortype is `str`

tockards:
This PR addresses issues in `get_sensor` in python 3 when no sensortype is supplied and the value is not a `str`.
It returns the value as quoted byte string. 
e.g if the value is a float that is `3.213` and the sensortype argument is `str` it will be retrieved from the katcp message arguments as `b'3.123'` then the `str` function will be applied to it and it will return as `"b'3.213'"`

This PR fixes that. 

related JIRA: https://skaafrica.atlassian.net/browse/P2M-37

### DIFF
--- a/katcp/testutils.py
+++ b/katcp/testutils.py
@@ -329,6 +329,9 @@ class BlockingTestClient(client.BlockingClient):
             if sensortype == bool:
                 typestr = "%r and then %r" % (int, bool)
                 value = bool(int(value))
+            elif sensortype == str:
+                typestr = "%r" % sensortype
+                value = bytes_to_native_str(value)
             else:
                 typestr = "%r" % sensortype
                 value = sensortype(value)


ajoubertza:
Thanks @tockards.  Fix looks OK, but we should really add a test that verifies this behaviour.  It was clearly missing from the test suite before.

katcp/test/test_compat.py
```diff
@@ -0,0 +1,20 @@
+# test_server.py
```
ajoubertza, line 1: Should this be `test_compat.py`?

katcp/test/test_compat.py
```diff
@@ -0,0 +1,20 @@
...
+# -*- coding: utf8 -*-
+# vim:fileencoding=utf8 ai ts=4 sts=4 et sw=4
+# Copyright 2009 National Research Foundation (South African Radio Astronomy Observatory)
```
ajoubertza, line 4: 2020 for the year - this file is new.

Jenkinsfile
```diff
@@ -79,11 +79,11 @@ pipeline {
...
-                        failNoReports: true,
-                        failUnhealthy: true,
-                        failUnstable: true,
-                        autoUpdateHealth: true,
-                        autoUpdateStability: true,
+                        //failNoReports: true,
+                        //failUnhealthy: true,
+                        //failUnstable: true,
+                        //autoUpdateHealth: true,
+                        //autoUpdateStability: true,
```
mmphego, line 86: I suppose we could remove these checks as it seems they aren't really as helpful.

katcp/test/test_compat.py
```diff
@@ -0,0 +1,20 @@
...
+
+import unittest
+
+from katcp.compat import ensure_native_str
+
+
+class test_CompatClass(unittest.TestCase):
+    def test_ensure_native_str(self):
+        with self.assertRaises(TypeError):
+            ensure_native_str(9)
```
mmphego, line 20: Perhaps also test for boolean, not sure if it's necessary though but assuming someone or thing passes a boolean - things would break...

mmphego (approved):
Looks good to me

PR #244, opened 2020-08-24
allow bytes and strings in test sensor list comparison of sensors.

tockards:
This PR adds a test and functionality that allows bytes and strings to work with `test_sensor_list`.

### DIFF
--- a/katcp/testutils.py
+++ b/katcp/testutils.py
@@ -642,9 +642,21 @@ class BlockingTestClient(client.BlockingClient):
         """
         def sensortuple(name, description, units, stype, *params):
             # ensure float params reduced to the same format
+            native_str_params = []
             if stype == "float":
                 params = ["%g" % float(p) for p in params]
-            return (name, description, units, stype) + tuple(params)
+            for param in params:
+                if is_bytes(param):
+                    str_param = ensure_native_str(param)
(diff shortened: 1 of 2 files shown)


ajoubertza (changes requested):
Approach looks good.  One formatting issue.

katcp/testutils.py
```diff
@@ -642,9 +642,21 @@ def test_sensor_list(self, expected_sensors, ignore_descriptions=False):
...
+                if is_bytes(param):
+                    str_param = ensure_native_str(param)
+                    native_str_params.append(str_param)
+                else:
+                    native_str_params.append(param)
+            return (
+                ensure_native_str(name),
+                ensure_native_str(description),
+                ensure_native_str(units),
+                ensure_native_str(stype),
```
mmphego, line 658: Would be nice to have a function that would take in the `name, description, units, stype,` and return the native str...But yeah looks good.

mmphego (approved):
Overall looks good to me.

ajoubertza (approved):
LGTM

katcp/testutils.py
tockards: thanks, I will apply this if we need it in other cases.

PR #253, opened 2020-11-04
address ensure_native_str returning newbytes

tockards:
There are instances when [newbytes](http://python-future.org/reference.html?highlight=newbytes) are supplied to `ensure_native_str` and a native str is not returned in python 2.

This PR addresses those issues.

### DIFF
--- a/katcp/compat.py
+++ b/katcp/compat.py
@@ -9,6 +9,8 @@
 from __future__ import absolute_import, division, print_function
 from future import standard_library
 standard_library.install_aliases()  # noqa: E402
+from future.types import newbytes
+from future.utils import bytes_to_native_str
 
 import builtins
 import future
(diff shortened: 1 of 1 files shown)


mmphego (approved):

ajoubertza:
@tockards Where are the `newbytes` coming from?  I'm not a fan of those data types as they lead to confusion.  I want to avoid them in any of CAM's code.  Can we just fix the usage instead?

tockards:
@ajoubertza it happens here parts is an array of `newbytes` in python 2
 https://github.com/ska-sa/katcp-python/blob/f7ece1ba46f39be730e98186edb9b5d4d5d27739/katcp/core.py#L600


this is where `name` is assigned to a `newbytes`
https://github.com/ska-sa/katcp-python/blob/f7ece1ba46f39be730e98186edb9b5d4d5d27739/katcp/core.py#L604


and this is where the `ensure_native_str` fails
https://github.com/ska-sa/katcp-python/blob/f7ece1ba46f39be730e98186edb9b5d4d5d27739/katcp/core.py#L616

tockards:
closing this PR as I addressed this.

PR #254, opened 2020-12-17
Add asyncio compatible ioloop to ioloop manager

tockards:
This PR aims to allow the ioloop_manager to also manage an asyncio loop on the tornado ioloop using the asyncio tornado bridge. 

see https://www.tornadoweb.org/en/branch4.5/_modules/tornado/platform/asyncio.html#AsyncIOLoop 

This is needed and can be used by python 3 packages that require the asyncio and not the tornado ioloop and you would like both to be managed by katcp-python.

Please note that katpc-python has a dependency on tornado 4.5.3 and therefore this is required.


katcp/ioloop_manager.py
```diff
@@ -52,7 +55,10 @@ def managed(self):
...
         if not self._ioloop:
             if self._ioloop_managed:
-                self.set_ioloop(tornado.ioloop.IOLoop())
+                if sys.version_info[0] == 3:
+                    self.set_ioloop(AsyncIOLoop())
+                else:
+                    self.set_ioloop(tornado.ioloop.IOLoop())
             else:
                 self.set_ioloop(tornado.ioloop.IOLoop.current())
         return self._ioloop
```
ajoubertza, line 64: Please add a test for this new functionality.
I tried this out in `test_ioloop_manager.py`'s `test_ThreadsafeMethodAttrWrapper` class.  It could be expanded, or we could even make a new test class for `IOLoopManager` that just tries the managed and unmanaged variations.

```python
    def test_managed_io_loop_is_asyncio_in_python3(self):
        if sys.version_info[0] == 3:
            self.assertTrue(hasattr(self.ioloop, "asyncio_loop"))
        else:
            self.assertFalse(hasattr(self.ioloop, "asyncio_loop"))
```

ajoubertza (changes requested):
OK, that should be OK for the managed IO loop.   Just need a test.

ajoubertza (approved):
LGTM

ajoubertza (approved):

tockards:
Tests are passing now and I am happy to go ahead with merging this and creating the new release.

PR #258, opened 2022-09-14
consistent error message in py2 and py3 for error in Timestamp decode

tockards:
This ensures that the error in Py2 and Py3 is consistent as the value is a `str`.

This is required for tests in other packages that rely on katcp-python and test katcp devices and failed responses.


lanceWilliams (approved):
Looks ok.

PR #259, opened 2022-09-15
only ensure_native_str on text.

tockards:
Only use `ensure_native_str` on text.

### DIFF
--- a/katcp/kattypes.py
+++ b/katcp/kattypes.py
@@ -420,7 +420,8 @@ class Timestamp(KatcpType):
         try:
             decoded = float(value)
         except:
-            value = ensure_native_str(value)
+            if is_text(value):
+                value = ensure_native_str(value)
             raise ValueError("Could not parse value '%s' as timestamp." %
                              value)
         if major < SEC_TS_KATCP_MAJOR:


katcp/kattypes.py
```diff
@@ -420,7 +420,8 @@ def decode(self, value, major):
         try:
             decoded = float(value)
         except:
-            value = ensure_native_str(value)
+            if is_text(value):
```
lanceWilliams, line 423: is there not maybe a version generic way to cast to string? Then we don't have to bother checking 'value' type before using it.

tockards: @lanceWilliams 
Not entirely sure on a generic way(need to do more research) as using `str` works but differs in python 2 and python 3. 

`ensure_native_str` is supposed to be our method but it fails when the value is not text.

I picked up the issue when I was fixing the tests for katsim work in python 2 and python 3 and the following failure was produced.
```python
Expected: reply('fail', "Error in parameter 1 (period): Could not parse value 'notanumber' as timestamp.")
Actual: reply('fail', "Error in parameter 1 (period): Could not parse value 'b'notanumber'' as timestamp.")
```

which is produced with the following line of code in kattypes. 
```python
raise ValueError("Could not parse value '%s' as timestamp." % value)
```

in python 2 it evaluates as 
```python
Python 2.7.18 (default, Jul  1 2022, 12:27:04) 
[GCC 9.4.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> value = 'notanumber'
>>> print("Could not parse value '%s' as timestamp." % value)
Could not parse value 'notanumber' as timestamp.
>>> value = b'notanumber'
>>> print("Could not parse value '%s' as timestamp." % value)
Could not parse value 'notanumber' as timestamp.
>>> 
```

And in python 3 it evaluates as
```python
Python 3.8.10 (default, Jun 22 2022, 20:18:18) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> value = 'notanumber'
>>> print("Could not parse value '%s' as timestamp." % value)
Could not parse value 'notanumber' as timestamp.
>>> value = b'notanumber'
>>> print("Could not parse value '%s' as timestamp." % value)
Could not parse value 'b'notanumber'' as timestamp.
```

I added the `is_text` part as there might be occurrences  when a value fails to be cast to a float and also fails to `ensure_native_str`.

tockards: We could do this before the `encode` and `decode` but that would would upend many of the python 2 and python 3 compatibility as when the project was started there was a decision to do this in the `encode` and `decode`

ludwigschwardt: I would rather ask why `ensure_native_str` is not working... It meant exactly to fix these kinds of problems:
```
In [6]: katcp.compat.ensure_native_str("notanumber")
Out[6]: 'notanumber'

In [7]: katcp.compat.ensure_native_str(b"notanumber")     # Python 3
Out[7]: 'notanumber'
```python

How does it manage to sneak in a `b` in front?

tockards: @ludwigschwardt thanks for your eyes and time. 
I plan to close this PR based on a discussion that I had with @lanceWilliams .
FYI the issue is in this test
https://github.com/ska-sa/katsim/blob/6a236f324ab482795c9448164ab75e8fde763faf/katsim/test/test_mkat_cbf_array_katcp.py#L281

